PR #9, opened 2016-07-06
Console

etiennebarrie:
This gives a way to play around with the gem objects in the dev environment.


es:
You're going to need to rebase against master and rename `recordstore` to `record_store` after https://github.com/Shopify/record_store/pull/9/files

es:
🚢 💨   after the rebase

PR #10, opened 2016-07-06
Handle zones as directories

etiennebarrie:
To make it easier to script committing changes to DNS records, we thought it would be easier if zones were broken up a bit into different files sitting in a directory.

I added support for records in separate files, but keeping all records in the zone file is still supported.

~~I moved to a directory with a file for the config and different files for the records.~~

* the zone file is still there, but records can be additionally stored in other files under a directory named after the zone (e.g. same as the zone file, except it's a directory and doesn't end in `.yml`)
* records are grouped into YAML files by FQDN and record type, which means there still can be multiple records inside a YAML file, but I think it makes sense
* no special case for wildcard domains, so there can be file names starting with `*`, I don't think it's a problem though.
* the previous layout is still supported, but zones are downloaded using the new layout
* I also added a way to "lint" a zone (all zones by default), which is somewhat similar to the existing `record-store sort` command, to help move from the single file zone to the new layout. LMK if you have a better name.
* ~~the config file is named `config.yml` but I'm thinking we should rename it `_config.yml` so that it sits at the top of the directory.~~


es:
You'll have to rebase this against master given the rename in https://github.com/Shopify/record_store/pull/12. Could you split the template and test zones so that one does a split of yaml/single record files, just a yaml file, and one that only uses single record files?

es:
I'm going to hold off on a more thorough review until the mix of directory/one file is implemented.

etiennebarrie:
Ok pushed new commits, instead of using either the new layout or the old one, we can now keep the zone file like before, but we can also add records in new files. Updated the description to match that.

We have now tests for:
* everything in the zone file: `zone-file.com`
* records in record files, only the config stays in the zone file: `one-record.com`, `two-records.com`
* mix of records in the zone file and in separate files: `merged-records.com`

es:
A lot of [these](https://github.com/Shopify/record_store/pull/10/files#diff-2d3e295619bf4546c351b2ac11565563R116) [functions](https://github.com/Shopify/record_store/pull/10/files#diff-ac4269553d420a5d50ffdbaf80b6fe13R58) [could](https://github.com/Shopify/record_store/pull/10/files#diff-ac4269553d420a5d50ffdbaf80b6fe13R81) use some spacing/newlines to make it easier to read.

es:
> no special case for wildcard domains, so there can be file names starting with *, I don't think it's a problem though.

 👍 

> records are grouped into YAML files by FQDN and record type, which means there still can be multiple records inside a YAML file, but I think it makes sense

:+1:

> the previous layout is still supported, but zones are downloaded using the new layout

Could we add a flag to the download command that'll allow you to specify what format you want the zone to be saved as? If write knows how to create zones in both formats, we should add that functionality to `reformat`/`lint`

etiennebarrie:
> Could we add a flag to the download command that'll allow you to specify what format you want the zone to be saved as? If write knows how to create zones in both formats, we should add that functionality to reformat/lint

How would you call that flag?

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
     validate :validate_provider_can_handle_zone_records
 
     def self.from_yaml_definition(name, definition)
```
es: Given it can be a single or multiple YAML definitions, it may make sense to rename this function to `load`.

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
...
+      definition['records'] ||= []
+      Dir["#{RecordStore.zones_path}/#{name}/*__*.yml"].each do |record_file|
```
es: I think this logic should live in the [`load_yaml_file` private method above](https://github.com/Shopify/record_store/blob/30247c9f706441fc454b9ecc61e957c225ea881b/lib/record_store/zone.rb#L29-L35). We can rename it to something more general.

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
...
+        provider: provider_name,
+        ignore_patterns: [{type: "NS", fqdn: "#{name}."}],
+      })
+    end
+
+    def self.write(name, config:, records:)
+      name = name.remove(/\.\z/)
+      zone_file = "#{RecordStore.zones_path}/#{name}.yml"
+      dir = "#{RecordStore.zones_path}/#{name}"
+      Dir["#{dir}/*"].each { |file| File.unlink(file) }
```
es: You can splat the return of Dir instead of having to iterate over the response

es: Now that I think about it, there's a couple problems with this ordering. We don't know if the directory actually exists and thus it'll error out before it can be created. It'd make more sense for the deleting of the files to be in `Errno::EEXIST`

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
...
+      begin
+        Dir.mkdir(dir)
+      rescue Errno::EEXIST
+      end
+      zone = { name => { config: config } }
+      File.write(zone_file, zone.deep_stringify_keys.to_yaml.gsub("---\n", ''))
```
es: We may want to extract the logic of writing a YAML file without the `---` into a helper

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
...
-        }
-      }.deep_stringify_keys.to_yaml.gsub("---\n", ''))
+      current_records = dns.retrieve_current_records.map(&:to_hash)
+      write(name, records: current_records, config: {
+        provider: provider_name,
+        ignore_patterns: [{type: "NS", fqdn: "#{name}."}],
+      })
+    end
+
+    def self.write(name, config:, records:)
```
es: What do you think if `Zone.write` were to take an instance of `Zone` as an argument or maybe even move `write` to be an instance method of `Zone`?

lib/record_store/zone.rb
```diff
@@ -106,6 +133,10 @@ def provider
       Provider.const_get(config.provider).new(zone: name.gsub(/\.\z/, ''))
     end
 
+    def write
+      self.class.write(name, config: config.to_hash, records: records.map(&:to_hash))
```
es: When will write be used without an instance of `Zone`? In other words, what's the reasoning behind extracting the values of a zone and using a static method of `write` instead of placing all of the logic here?

es:
I'm assuming that an API server will be written above this to allow for automated DNS updates. How do you see creating/updating/deleting APIs working with the current write API? Will the service load the entire zone and make updates to the Zone in memory? If so, that logic is probably best encompassed in Record Store.

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
...
+      File.write(zone_file, zone.deep_stringify_keys.to_yaml.gsub("---\n", ''))
+      records.group_by { |record| [record.fetch(:fqdn), record.fetch(:type)] }.each do |(fqdn, type), grouped_records|
+        grouped_records.sort_by! {|record| record[:nsdname] || record[:address] }
+        grouped_records.each do |record|
+          record.delete(:fqdn)
+          record.delete(:type)
+          record.deep_stringify_keys!
+        end
+        grouped_records = grouped_records.first if grouped_records.size == 1
+        File.write("#{dir}/#{fqdn.remove(/\.\z/)}__#{type}.yml", grouped_records.to_yaml.gsub("---\n", ''))
```
es: As implied by an earlier comment, `Zone#write`/`Zone.write` should be able to save the zone in both formats (single yaml file and directory of records).

lib/record_store/zone.rb
etiennebarrie: When you download, you have a bunch of data without `Zone` objects.

lib/record_store/zone.rb
etiennebarrie: No Dir[] will return an empty array if there's no file, and it won't do a thing? But in my updated commits, I always remove the dir and re-create it after if the format is directory.

lib/record_store/zone.rb
```diff
@@ -56,22 +56,49 @@ def yaml_files
     validate :validate_provider_can_handle_zone_records
 
     def self.from_yaml_definition(name, definition)
+      definition['records'] ||= []
+      Dir["#{RecordStore.zones_path}/#{name}/*__*.yml"].each do |record_file|
+        fqdn, type = File.basename(record_file, '.yml').split('__')
```
es: Feel free to ignore if you think I'm bikeshedding, but I'd prefer for the format to be `RR_TYPE__LABEL.yml` (e.g. `A__a-record.dnsimple.example.com.yml`.

None 🚲 🏠 : since we have the current origin in the directory name, we can shorten the name to omit the origin. Example: an A record for `test.example.com` could be `example.com/A__test.yml`.

lib/record_store/zone.rb
etiennebarrie: I'm going to keep all the YAML parsing in `YamlDefinitions`, it's true that it leaked there.

lib/record_store/zone.rb
es: The guarantee we're making then is if you're using a directory, you could lose all your comments, but if you're using a single yaml file they'll be preserved? Which I think is 💯

lib/record_store/zone.rb
es: Maybe we should change the API for `Zone.download` to return a `Zone` instance?

lib/record_store/zone.rb
etiennebarrie: The comments are lost in any case if you're reformating.

lib/record_store/zone.rb
es: Hrm, for some reason it's not writing to the zone paths (at least from the looks of it given: [`template/dnsimple.example.com/config.yml`](https://github.com/Shopify/record_store/blob/30247c9f706441fc454b9ecc61e957c225ea881b/template/dnsimple.example.com/config.yml) - it should be `template/zones/dnsimple.example.com/config.yml`)

test/zone_test.rb
```diff
@@ -211,14 +228,14 @@ def test_filter_records
     assert_equal 1, Zone.filter_records(records, [{cname: 'real.example.com.' }]).length
   end
 
-  def test_download_downloads_zone_into_file
+  def test_download_downloads_zone_into_dir
```
es: You'll need to add a duplicate of this test for both formats

es:
Can one of the test records be a wildcard on the origin to confirm that works?

etiennebarrie:
> I'm assuming that an API server will be written above this to allow for automated DNS updates. How do you see creating/updating/deleting APIs working with the current write API? Will the service load the entire zone and make updates to the Zone in memory? If so, that logic is probably best encompassed in Record Store.

So the idea here is mostly to have a file layout that is easier to work with (e.g. adding/removing a file rather than adding/removing content to a YAML file).
My plan is to wrap the GitHub API to do those changes, but indeed having a fully-fledge API here would be nice too. I don't see those two goals as incompatible though, but this PR focuses on that part.

etiennebarrie:
Ok I think I'm done based on your feedback:
* switched to RTYPE__DOMAIN for the file name
* added a test with a wildcard domain and the file including `*`
* moved all the YAML-related logic in the `YamlDefinitions` module 
* extracted the logic of writing a YAML file without the --- into a helper

lib/record_store/cli.rb
```diff
@@ -107,10 +109,27 @@ def download
...
 
       puts "Downloading records for #{name}"
-      Zone.download(name, provider)
+      Zone.download(name, provider, format: format)
       puts "Records have been downloaded & can be found in zones/#{name}.yml"
     end
 
+    option :name, desc: 'Zone to reformat', aliases: '-n', type: :string, required: false
+    option :format, desc: 'Format (file or directory)', aliases: '-f', type: :string, default: 'file'
+    desc 'reformat', 'Sorts and re-outputs the zone (or all zones) as directory'
```
es: `as directory`/`as specified format`

lib/record_store/cli.rb
```diff
@@ -251,5 +270,10 @@ def zones_modified
 
       modified_zones
     end
+
+    def validate_format
```
es: `validate_format` should have a different name since it's returning the format.

lib/record_store/zone.rb
```diff
@@ -21,19 +21,74 @@ def each(&block)
 
       def find(name)
         return unless File.exists?(zone_path = "#{RecordStore.zones_path}/#{name}.yml")
-        Zone.from_yaml_definition(*YAML.load_file(zone_path).first)
+        load_yml_zone_definition(zone_path).first.last
```
es: We can cleanup this method to be just `defined[name]`, but it's a bit of a yakshave to do without refactoring the stubs in `ZoneTest` (took me a while to dig down). We can do that in a follow-up PR.

lib/record_store/zone.rb
```diff
@@ -21,19 +21,74 @@ def each(&block)
...
+      def write_yml_file(filename, data)
+        lines = data.deep_stringify_keys.to_yaml.lines
+        lines.shift if lines.first == "---\n"
+        File.write(filename, lines.join)
+      end
+
+      def load_yml_zone_definition(filename)
         result = {}
+        dir = File.dirname(filename)
         YAML.load_file(filename).each do |name, definition|
```
es: By swapping the `each` to a `first` we can unroll this loop making it clearer how this works.

lib/record_store/zone.rb
```diff
@@ -21,19 +21,74 @@ def each(&block)
...
+          definition['records'] ||= []
+          Dir["#{dir}/#{name}/*__*.yml"].each do |record_file|
+            definition['records'] += load_yml_record_definitions(name, record_file)
+          end
           result[name] = Zone.from_yaml_definition(name, definition)
         end
         result
```
es: With the loop unrolled, we can inline the hash definition or move all hash logic into `Zone.defined`

lib/record_store/zone.rb
```diff
@@ -21,19 +21,74 @@ def each(&block)
...
       end
 
+      def load_yml_record_definitions(name, record_file)
+        type, domain = File.basename(record_file, '.yml').split('__')
+        records_definitions = Array.wrap(YAML.load_file(record_file))
+        records_definitions.each do |record_definition|
```
es: What do you think about changing `each` to `map` and merging the `fqdn` and `type` into the record definition?

lib/record_store/zone.rb
```diff
@@ -21,19 +21,74 @@ def each(&block)
...
+      end
+
+      def write(name, config:, records:, format: :file)
+        raise ArgumentError, "format must be :directory or :file" unless %i(file directory).include?(format)
+        name = name.chomp('.')
+        zone_file = "#{RecordStore.zones_path}/#{name}.yml"
+        zone = { name => { config: config.to_hash } }
+        records = records.map(&:to_hash).sort_by! {|r| [r.fetch(:fqdn), r.fetch(:type), r[:nsdname] || r[:address]]}
+
+        zone[name][:records] = records if format == :file
```
es: Since `format` can only ever be one value, I think `if format == :file` and `if format == :directory` should be one `if/else` block.

lib/record_store/zone.rb
```diff
@@ -21,19 +21,74 @@ def each(&block)
...
+      def write_record_files(name, records)
+        dir = "#{RecordStore.zones_path}/#{name}"
+        Dir.mkdir(dir)
+        records.group_by { |record| [record.fetch(:fqdn), record.fetch(:type)] }.each do |(fqdn, type), grouped_records|
+          grouped_records.each do |record|
+            record.delete(:fqdn)
+            record.delete(:type)
+            record.deep_stringify_keys!
+          end
+          grouped_records = grouped_records.first if grouped_records.size == 1
```
es: Why do we need this?

es:
@wvanbergen can you please take a look at this PR as well? Given the size of the change, I'd like 2 👀 before 🚀

lib/record_store/zone.rb
etiennebarrie: What do you mean by `unroll`? If we use `first` over `each`, that means only one zone can be defined in a file. Right now the code handles the case where there are multiple zones in a single file, as long as you access it through `defined` and not `find`.
There are no tests for this case though and this PR keeps the existing behaviour even though it's untested.

lib/record_store/cli.rb
etiennebarrie: Right, I found the `enum` option of Thor which should make things better.

lib/record_store/zone.rb
```diff
@@ -21,19 +21,72 @@ def each(&block)
...
```
etiennebarrie: This, combined with the `Array.wrap` call in `load_yml_record_definitions ` allows to have a simpler YAML file:
```yaml
ttl: 86400
address: 10.10.10.42
```

Rather than:
```yaml
- ttl: 86400
  address: 10.10.10.42
```

I'm no longer sure it's worth doing that though, we should probably have a single format for record files. But for example for CNAMEs where it's not legal to have multiple records, it's weird to force people to use a YAML array with a hash inside rather than a straight hash. WDYT?

lib/record_store/zone.rb
etiennebarrie: I did that in bbf375156d23fa3a52f245a0eebc1f122485cf3d but I'm not sure I want to keep it, because some of the logic is shared.
```ruby
if format == :file
  zone[name][:records] = records
  write_yml_file(zone_file, zone.deep_stringify_keys)
  remove_record_files(name)
else
  write_yml_file(zone_file, zone.deep_stringify_keys)
  remove_record_files(name)
  write_record_files(name, records)
end
```

etiennebarrie:
Rebased and squashed fixes from before, added new commits for the new feedback.

lib/record_store/zone.rb
es: I think the mixed format makes sense. I'm sure a lot of people are going to run into not formatting it as an array otherwise for the exact reason you mentioned.

lib/record_store/zone.rb
es: I don't see a reason to support the ability to declare multiple zones in a single file. I'm okay with 🔥  the untested functionality.

test/zone_test.rb
```diff
@@ -8,6 +8,23 @@ def test_find_returns_zone_by_name
     assert_predicate zone, :valid?
   end
 
+  def test_find_returns_zone_by_name_in_single_file
+    zone = Zone.find('zone-file.com')
+
+    assert_equal Zone, zone.class
```
es: `assert_instance_of`?

lib/record_store/zone.rb
etiennebarrie: I'll do that in a follow-up PR, along with having `find` just go through `defined`,  if that's ok.

es:
🚢 💨 

@wvanbergen can you give this a second set of 👀 ?

es:
Ping @wvanbergen

lib/record_store/zone.rb
```diff
@@ -21,19 +21,77 @@ def each(&block)
...
+        end
       end
 
       private
 
-      def load_yaml_file(filename)
+      def write_yml_file(filename, data)
+        lines = data.to_yaml.lines
+        lines.shift if lines.first == "---\n"
+        File.write(filename, lines.join)
```
wvanbergen: Maybe `data.to_yaml.sub(\A\-\-\-$, '')` so it's a oneliner? YMMV.

test/zone_test.rb
```diff
@@ -251,6 +268,72 @@ def test_zone_unchanged_describes_if_zone_matches_provider
...
+      type: 'CNAME',
+      fqdn: '*.wildcard.com',
+      cname: 'wildcard.com',
+      ttl: 60,
+    }])
+
+    zone.write(format: :directory)
+
+    assert Dir.exist?("#{RecordStore.zones_path}/wildcard.com")
+    assert File.exist?("#{RecordStore.zones_path}/wildcard.com/CNAME__*.yml")
```
wvanbergen: Is the usage of an asterisk in a filename going to super annoying when trying to navigate this folder using a shell?

wvanbergen:
:+1:. Nice. we've been talking about this one for a while. Glad to see it is happening!

test/zone_test.rb
etiennebarrie: I don't think so, bash deals with it pretty well by default.

PR #15, opened 2016-08-25
[WIP] Adding ALIAS record support for DynECT in Record Store

andrewlouis93:
r/ @es

- This PR allows DynECT zones to use ALIAS dns entries.
- It also uses a vendored version of the `fog-dynect` gem created here: https://github.com/Shopify/fog-dynect that allows ALIAS records to be used when interfacing with DynECT.

Related: https://github.com/Shopify/growth_engineering/issues/219

Note: Hasn't been 🎩-ted yet, our zones need to be granted the ability to use ALIAS records first.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'ejson'
 gem 'fog'
 gem 'fog-json'
 gem 'fog-xml'
-gem 'fog-dynect'
+gem 'fog-dynect', :git => 'https://github.com/Shopify/fog-dynect.git', :branch => 'alias-record-support'
 
 group :test do
   gem 'mocha'
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'ejson'
 gem 'fog'
 gem 'fog-json'
 gem 'fog-xml'
-gem 'fog-dynect', :git => 'https://github.com/Shopify/fog-dynect.git'
+gem 'fog-dynect', :git => 'https://github.com/fog/fog-dynect.git'
 
 group :test do
   gem 'mocha'
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'ejson'
 gem 'fog'
 gem 'fog-json'
 gem 'fog-xml'
-gem 'fog-dynect', :git => 'https://github.com/Shopify/fog-dynect.git', :branch => 'alias-record-support'
+gem 'fog-dynect', :git => 'https://github.com/Shopify/fog-dynect.git'
 
 group :test do
   gem 'mocha'
--- a/lib/record_store/provider/dynect.rb
+++ b/lib/record_store/provider/dynect.rb
@@ -2,6 +2,10 @@ require 'fog/dynect'
 
 module RecordStore
   class Provider::DynECT < Provider
+    def self.record_types
+      super.add('ALIAS')
+    end
+
     def freeze_zone
       session.put_zone(@zone_name, freeze: true)
     end
--- a/test/providers/dynect_test.rb
+++ b/test/providers/dynect_test.rb
@@ -173,8 +173,8 @@ class DynECTTest < Minitest::Test
 
   def test_apply_changeset_sets_state_to_match_changeset
     a_record = Record::ALIAS.new(
-      fqdn: 'dns-test-dyn.shopify.io',
-      zone: 'shopify.io',
+      fqdn: 'alias-test.dns-test-dyn.shopify.io',
+      zone: 'dns-test-dyn.shopify.io',
       ttl: 86400,
       alias: 'meaniepies.myshopify.com'
     )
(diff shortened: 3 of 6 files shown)


Gemfile
```diff
@@ -8,7 +8,7 @@ gem 'ejson'
 gem 'fog'
 gem 'fog-json'
 gem 'fog-xml'
-gem 'fog-dynect'
+gem 'fog-dynect', :git => 'https://github.com/Shopify/fog-dynect.git'
```
es: Is there a reason we didn't PR the change to upstream? I'd be very hesitant to start maintaining a fork.

andrewlouis93: @es since by default, ALIAS records are not configured - I was under the assumption that it was a deliberate choice by the fog-dynect developers we were working around.

I've just created a PR in any case: https://github.com/fog/fog-dynect/pull/6

es: It's documented on the [DynECT website](http://dyn.com/managed-dns/alias/). I think no one just added it to the library.

andrewlouis93:
@es I've gotten the integration test working (looks good!) and the PR to fog-dynect successfully merged in; see: https://github.com/fog/fog-dynect/pull/6

Could you give it one more review?

sbfaulkner:
since they've merged your stuff upstream, probably burn the Shopify fork and point at the upstream instead???

andrewlouis93:
done, and done :)

lib/record_store/provider/dynect.rb
```diff
@@ -2,6 +2,10 @@
 
 module RecordStore
   class Provider::DynECT < Provider
+    def self.record_types
+      super.add('ALIAS')
```
sbfaulkner: @es do you think we should keep this as a specialization, or just add ALIAS to the collection of record_types in the base Provider class (now that both providers support it)?

sbfaulkner: ie. and remove the override in both Provider::DynECT and Provider::DNSimple

es: We need to figure out a way for Record Store to be aware of a zone having the ability to create ALIAS records. Otherwise, people are going to get confusing error messages when attempting to add ALIAS records to unsupported zones.

test/zone_test.rb
```diff
@@ -47,14 +47,6 @@ def test_zone_has_config_with_provider
     assert_instance_of Zone::Config, zone.config
   end
 
-  def test_zone_is_invalid_if_provider_does_not_support_record_type
-    zone = Zone.find('empty.com')
-    assert_equal 'DynECT', zone.config.provider
-    zone.records = [Record::ALIAS.new(fqdn: zone.name, ttl: 60, cname: "alias.#{zone.name}")]
-
-    refute_predicate zone, :valid?
-  end
```
es: I'd like to keep the test of this logic somehow, but we can't use ALIAS anymore for testing 😕

test/providers/dynect_test.rb
```diff
@@ -40,6 +40,23 @@ def test_build_aaaa_from_api
...
   end
 
+  def test_build_alias_from_api
+    record = @dyn.send(:build_from_api, {
+      "zone" => 'dns-test.shopify.io',
+      "ttl" => 60,
+      "fqdn" => "alias.dns-test.shopify.io",
+      "record_type" => "ALIAS",
+      "rdata" => {
+        "cname" => "dns-test.shopify.io."
```
es: Unless I'm mistaken, this should be `alias`

es:
Have you validated this works by using your local development version to create/update/delete an ALIAS record for a DynECT zone?

test/providers/dynect_test.rb
andrewlouis93: @es what were you referring to here? It does say ALIAS.

andrewlouis93:
@es no I haven't, will write tests for this and ping.

test/providers/dynect_test.rb
es: `cname`

andrewlouis93: @es oh, that's correct. See here: https://github.com/Shopify/record_store/blob/master/lib/record_store/record/alias.rb#L5

es: Right, but the [API docs specify](https://help.dyn.com/create-alias-records-api/) the attribute `alias`. This code will break right now if it tries to interact with DynECT's API.

andrewlouis93: Got it, fixed.

lib/record_store/provider/dynect.rb
sbfaulkner: @es re: this comment ... I'm really not sure how we can tell that the zone supports ALIAS records (in part because it also requires explicit permission to manage them - not just that the zone has to "support" them)

did you have any idea(s) for this?

sbfaulkner:
closing in favour of #17

PR #16, opened 2016-09-15
Validates that all records with matching FQDN and type share the same TTL

### LINKED ISSUE #5: Ensure all {fqdn,type} records have the same TTL
From DynECT technical support:
```
The use of differing TTL values for records within a record set has been deprecated as of RFC-2181. This means that for a given hostname, all the records of a given record type will need to use the same TTL value.
```

A validation should be added to zones ensure this spec is adhered to.

wvanbergen:
I am not sure if this is part of the DNS standard, or whether this is a limitation of just DynECT. But I think it makes sense to enforce this anyway.

Fixes #5. For review: @es and @andrewlouis93

### DIFF
--- a/lib/record_store/zone.rb
+++ b/lib/record_store/zone.rb
@@ -108,7 +108,6 @@ module RecordStore
     validate :validate_records
     validate :validate_config
     validate :validate_all_records_are_unique
-    validate :validate_a_records_for_same_fqdn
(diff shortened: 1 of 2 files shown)


wvanbergen:
I noticed we have a validation for matching TTLs on just A records already. Should I just remove that?

wvanbergen:
New pair of 👀 please!

PR #17, opened 2016-09-19
Enable DynECT ALIAS support

sbfaulkner:
take 2

@es if you can please take a look... especially interested in you thoughts on tests... are we missing anything here?

/cc @andrewlouis93


lib/record_store/provider/dynect.rb
```diff
@@ -83,15 +87,21 @@ def secrets
...
+      record_type = api_record.fetch('record_type')
       record = api_record.merge(api_record.fetch('rdata')).slice!('rdata').symbolize_keys
 
-      return if record.fetch(:record_type) == 'SOA'
+      return if record_type == 'SOA'
+
+      case record_type
+      when 'ALIAS'
+        record[:cname] = record.delete(:alias)
+      end
```
andrewlouis93: @sbfaulkner @es I'm thinking we can eliminate this and modify record/alias.rb to use an attribute called `alias` instead of `cname` for consistency.

DNSimple doesn't seem to rely on the property name `cname` with ALIAS records.  Doing this would ensure that we have a consistent naming pattern in managing records, instead of stubbing Record objects after they've been created.

dev.yml
```diff
@@ -5,6 +5,9 @@ up:
   - ruby: 2.3.1
   - bundler
 
+commands:
+  test:
```
es: What do you think about including the ability to specify a test file like we do in [nginx-routing-modules](https://github.com/Shopify/nginx-routing-modules/blob/1962ec1fc78b2621ea0f122031819a9d54ec9882/dev.yml#L38-L48)?

lib/record_store/provider/dynect.rb
es: 100% agree.

test/zone_test.rb
```diff
@@ -48,6 +48,9 @@ def test_zone_has_config_with_provider
   end
 
   def test_zone_is_invalid_if_provider_does_not_support_record_type
+    supported_record_types = Provider::DynECT.record_types - ['ALIAS']
+    Provider::DynECT.stubs(:record_types).returns(supported_record_types)
```
es: We need to think through how we want to model this since some zones can't support `ALIAS` records (💩  DynECT 💩 ). 

I'm currently leaning to making it an explicit config option for zones that belong to DynECT and adding a validation onto the zone. The alternative is we need to make sure there's a test that can run during CI to catch `ALIAS` records being added to unsupported zones.

lib/record_store/provider/dynect.rb
andrewlouis93: @es Thanks for your note! I just realized, although this would require: 
- In record-store, change all DNSimple ALIAS records' `cname` attributes to `alias`, OR: 
- (Dirty, but minimal impact) Make DNSimple read the `cname` attribute if the `alias` attribute is empty here: https://github.com/Shopify/record_store/blob/master/lib/record_store/provider/dnsimple.rb#L138

what do you think?

test/zone_test.rb
andrewlouis93: Maybe I'm not understanding your primary alternative properly, what would this explicit config option be? And, how is that option preferable to a simple CI test to check that the user is not attempting to add ALIAS records to an unsupported zone?

sbfaulkner: it doesn't actually HAVE to be an apex... you CAN define an ALIAS that is NOT an apex

dev.yml
sbfaulkner: done

lib/record_store/provider/dynect.rb
es: Breaking changes are fine in my opinion. We just need to increment a major version when cutting a new package.

test/zone_test.rb
es: @andrewlouis93 only DynECT zones would be able to specify they accept ALIAS records in [the config block](https://github.com/Shopify/record_store/blob/368153106068800325b4e579483faa427afe7add/template/zones/dnsimple.example.com.yml#L3). You'd have a new possible config along the lines of `supports_alias: {true|false}` (we can set a default of false if missing).

lib/record_store/record.rb
```diff
@@ -18,6 +18,12 @@ def initialize(record)
     end
 
     def self.build_from_yaml_definition(yaml_definition)
+      case yaml_definition.fetch(:type)
```
sbfaulkner: as discussed... let's put the type in to a variable  that we then reference here, and below...

sbfaulkner:
@andrewlouis93 most recent commits begin the testing of the config and the validation of the provider's alias support and that the alias is an apex/root

lib/record_store/provider/dnsimple.rb
```diff
@@ -19,7 +19,7 @@ def add(record)
 
       if record.type == 'ALIAS'
         txt_alias = retrieve_current_records.detect do |rr|
-          rr.type == 'TXT' && rr.fqdn == record.fqdn && rr.txtdata == "ALIAS for #{record.cname.gsub(/.\z/, '')}"
+          rr.type == 'TXT' && rr.fqdn == record.fqdn && rr.txtdata == "ALIAS for #{record.alias.gsub(/.\z/, '')}"
```
sbfaulkner: this pattern actually removes the last character... not just a trailing `.`

we should escape the `.` as well

sbfaulkner:
I think this almost does it... the one case that I think is not yet handled is that existing DNSimple zone files would act as if `supports_alias: false` were in the config.

I think we need to move the reading of the provider support value (ie. the default) into the config object... alternately we need to set up the default when reading.

TBD tomorrow

lib/record_store/zone/config.rb
```diff
@@ -3,13 +3,16 @@ class Zone
...
+      alias supports_alias? supports_alias
 
       validate :validate_zone_config
 
-      def initialize(ignore_patterns: [], provider: nil)
+      def initialize(ignore_patterns: [], provider: nil, supports_alias: nil)
         @ignore_patterns = ignore_patterns
         @provider = provider
+        @supports_alias = supports_alias
+        @supports_alias = Provider.const_get(provider).supports_alias? if @supports_alias.nil? && valid_provider?
```
sbfaulkner: not 100% happy about putting this here, since it means that Config is no longer as naive as it was, but it gets ugly fast when we try to move it up a level

test/zone/config_test.rb
```diff
@@ -32,4 +32,12 @@ def test_config_is_properly_loaded_from_zonefile
...
     assert_equal 'DynECT', zone.config.provider
   end
+
+  def test_config_supports_alias_based_on_provider
+    config = Zone.new('dynect-config.com', config: {provider: 'DynECT'}).config
+    refute_predicate config, :supports_alias?
+
+    config = Zone.new('dnsimple-config.com', config: {provider: 'DNSimple'}).config
+    assert_predicate config, :supports_alias?
+  end
```
es: It might be a little overkill, but I'd add a sanity check test that `supports_alias` is an accepted config.

test/providers/dynect_test.rb
```diff
@@ -6,6 +6,10 @@ def setup
     @dyn = Provider::DynECT.new(zone: @zone_name)
   end
 
+  def test_supports_alias_by_default
+    refute Provider::DynECT.supports_alias?, 'DynECT should NOT support ALIAS by default'
```
andrewlouis93: No predicate here though, `Provider::DynECT.supports_alias?` is a Boolean, same with the previous comment.

lib/record_store/zone/config.rb
andrewlouis93: I'm okay with keeping it as is - if we move the logic into a public method, we wouldn't be able to inspect the supports_alias value after the Config object is initialized like so: https://github.com/Shopify/record_store/blob/f0f599d7e21ba2cbfd4551db135a12df9d388bf4/test/zone_test.rb#L405

es: I'm not seeing `supports_alias` inspected anywhere in that unit test

lib/record_store/zone/config.rb
```diff
@@ -4,15 +4,18 @@ class Config
...
       validate :validate_zone_config
 
       def initialize(ignore_patterns: [], provider: nil, supports_alias: nil)
         @ignore_patterns = ignore_patterns
         @provider = provider
         @supports_alias = supports_alias
+      end
+
+      def supports_alias?
         @supports_alias = Provider.const_get(provider).supports_alias? if @supports_alias.nil? && valid_provider?
```
es: Since this is an reader, it shouldn't be modifying any internal state. How about `@supports_alias || (valid_provider? && Provider.const_get(provider).supports_alias?)`?

sbfaulkner: that ^ doesn't *quite* work, since an explicit `false` in the config would be overridden by the default provider value...

it could be...

```
def supports_alias?
  @supports_alias.nil? && valid_provider? ? Provider.const_get(provider).supports_alias? : @supports_alias
end
```

test/providers/dynect_test.rb
sbfaulkner: @andrewlouis93 

`refute_predicate Provider::DynECT, :supports_alias?`

test/providers/dnsimple_test.rb
```diff
@@ -6,6 +6,10 @@ def setup
     @dnsimple = Provider::DNSimple.new(zone: @zone_name)
   end
 
+  def test_supports_alias_by_default
+    assert Provider::DNSimple.supports_alias?, 'DNSimple should support ALIAS by default'
```
sbfaulkner: @andrewlouis93 
`assert_predicate Provider::DNSimple, :supports_alias?`

test/zone/config_test.rb
sbfaulkner: @andrewlouis93 

emil is suggesting we add config like this... https://github.com/Shopify/record_store/blob/34a02f9f6d5972039c8dd0b8e165fed31fdb4bf8/test/zone/config_test.rb#L25

and test for the value being set similar to this...
https://github.com/Shopify/record_store/blob/34a02f9f6d5972039c8dd0b8e165fed31fdb4bf8/test/zone/config_test.rb#L33

README.md
```diff
@@ -48,9 +48,11 @@ The permissions required are broken into 2 groups:
 * WRITE: `RecordAdd`, `RecordDelete`, `ZonePublish`, `ZoneDiscardChangeset`, `ZoneFreeze`, `ZoneThaw`, `ZoneAddNode`,
 `ZoneRemoveNode`
 
-All CI validations only require READ permissions; deplyoing requires a user with READ and WRITE permissions.
+All CI validations only require READ permissions; deploying requires a user with READ and WRITE permissions.
 
-For a breakdown of what each permssion allows read through [DynECT's permissions guide](https://help.dyn.com/user-and-group-permissions/).
+In addition, the `AliasService` permission is required to be able to read or write ALIAS records on DynECT.
```
sbfaulkner: there were two reasons... 1. this 2. the api version used (the fog-dynect gem used an old api version that didn't support ALIAS - ie. dynect wouldn't accept OR return ALIAS records)

README.md
```diff
@@ -145,7 +147,7 @@ end
 
 #### Provider-Specific Records
 
-For provider-specific records (e.g. `ALIAS`), create the record model in `lib/record_store/record` as any other record. In the provider, extend `self.record_types` and append the custom record types to the `Set` returned by `Provider.record_types` (e.g. [`DNSimple.record_types`](https://github.com/Shopify/record_store/blob/1ec0d1410cf8bedf79bc63e8e4cdc7cdb0f1019b/lib/record_store/provider/dnsimple.rb#L5-L7)).
+For provider-specific records (e.g. `ALIAS`), create the record model in `lib/record_store/record` as any other record. In the provider, extend `self.record_types` and append the custom record types to the `Set` returned by `Provider.record_types` (e.g. [`DNSimple.record_types`](https://github.com/Shopify/record_store/blob/e8fb991ae827afb932f97930b55f046aa463d58f/lib/recordstore/provider/dnsimple.rb#L6)).
```
sbfaulkner: @andrewlouis93 https://github.com/Shopify/record_store/blob/master/lib/record_store/provider/dnsimple.rb#L6 should be the correct URL here

test/providers/dynect_test.rb
andrewlouis93: 🙈

README.md
andrewlouis93: @sbfaulkner @es hmm I'm guessing we should rewrite this comment completely. After we merge this PR in, that comment wouldn't make sense, it would point at here: https://github.com/Shopify/record_store/blob/58e8f525500e2d43c8546cccf7e09b56b1bc0a3a/lib/record_store/provider/dnsimple.rb#L5

Should we rewrite that comment completely? Or eliminate it? I don't know if there are additional major record types with questionable support outside of ALIAS records.

sbfaulkner:
@andrewlouis93 think you misse this one... https://github.com/Shopify/record_store/pull/17#discussion_r79896249

PR #18, opened 2016-09-19
Make Record::ALIAS use the alias attribute

andrewlouis93:
Makes Record::ALIAS use the alias attribute instead of the cname attribute. 

@es @sbfaulkner

### DIFF
--- a/lib/record_store/record.rb
+++ b/lib/record_store/record.rb
@@ -18,6 +18,12 @@ module RecordStore
     end
 
     def self.build_from_yaml_definition(yaml_definition)
+      case yaml_definition.fetch(:type)
+      when 'ALIAS'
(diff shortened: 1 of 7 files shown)


lib/record_store/record.rb
```diff
@@ -18,6 +18,12 @@ def initialize(record)
     end
 
     def self.build_from_yaml_definition(yaml_definition)
+      case yaml_definition.fetch(:type)
+      when 'ALIAS'
+        if yaml_definition.key?(:cname)
+          yaml_definition[:alias] = yaml_definition.delete(:cname)
+        end
+      end
```
sbfaulkner: I think I like this for a safety precaution... we can remove it after shipping for all the existing zones

PR #19, opened 2016-10-26
Support multiple providers for individual zones

es:
For: https://github.com/Shopify/service-disruptions/issues/530, https://github.com/Shopify/traffic/issues/998, https://github.com/Shopify/traffic/issues/169 

**Problem**:
Record Store currently only supports a single provider. This requires us to manually change the provider of every single zone and deploy.

**Solution**:
By deploying to multiple providers at all times, we can have a zone's NS records point to more than one (lowering the impact of a DNS provider outage). This PR adds the ability to specify multiple providers for a single zone. 

Before a zone would specify:
```
provider: DynECT
```

Now the config *must* be changed to an array format:
```
providers:
  - DynECT
```

To have the zone synced to multiple providers at the same time, simply specify an additional provider in the array:
```
providers:
  - DynECT
  - DNSimple
```

**How?**
The main change was disassociating providers from individual zones. Now providers are static classes. Changesets have been modified to be provider specific. They have a static method to create a changeset between a zone and a specific provider. Calling `Changeset#apply` will apply the `Change`s of the changeset to synchronize the state. `Zone#build_changesets` will build a changeset for every provider of a zone.

Paired with @klautcomputing 

r: @wvanbergen @etiennebarrie @sbfaulkner @andrewlouis93 @marc-barry @camilo

### DIFF
--- a/lib/record_store/zone/yaml_definitions.rb
+++ b/lib/record_store/zone/yaml_definitions.rb
@@ -61,16 +61,19 @@ module RecordStore
         raise 'more than one zone in file' if data.size > 1
         name, definition = data.first
         definition['records'] ||= []
+        definition['records'] = definition['records'].map {|hash| hash.inject({}){|h,(k,v)| h[k.to_sym] = v; h}}
         Dir["#{dir}/#{name}/*__*.yml"].each do |record_file|
           definition['records'] += load_yml_record_definitions(name, record_file)
         end
-        Zone.new(name, definition.deep_symbolize_keys)
+        Zone.new(name: name, records: definition['records'], config: definition['config'])
       end
 
       def load_yml_record_definitions(name, record_file)
         type, domain = File.basename(record_file, '.yml').split('__')
         Array.wrap(YAML.load_file(record_file)).map do |record_definition|
-          record_definition.merge(fqdn: "#{domain}.#{name}", type: type)
+          record_definition.merge('fqdn' => "#{domain}.#{name}", 'type' => type)
+        end.map do |hash|
+          hash.inject({}){|h,(k,v)| h[k.to_sym] = v; h}
         end
       end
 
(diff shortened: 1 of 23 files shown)


lib/record_store/zone/yaml_definitions.rb
```diff
@@ -61,16 +61,19 @@ def load_yml_zone_definition(filename)
...
+        Zone.new(name: name, records: definition['records'], config: definition['config'])
       end
 
       def load_yml_record_definitions(name, record_file)
         type, domain = File.basename(record_file, '.yml').split('__')
         Array.wrap(YAML.load_file(record_file)).map do |record_definition|
-          record_definition.merge(fqdn: "#{domain}.#{name}", type: type)
+          record_definition.merge('fqdn' => "#{domain}.#{name}", 'type' => type)
+        end.map do |hash|
+          hash.inject({}){|h,(k,v)| h[k.to_sym] = v; h}
```
etiennebarrie: We could use `symbolize_keys` here.

test/test_helper.rb
```diff
@@ -13,6 +13,11 @@ class Minitest::Test
 
   RecordStore.config_path = DUMMY_CONFIG_PATH
 
+  def teardown
+    Provider::DynECT.instance_variable_set(:@dns, nil)
+    Provider::DNSimple.instance_variable_set(:@dns, nil)
```
etiennebarrie: Maybe we need a `clear_session` or something similar?

es: I don't think `clear_session` is needed since we want to cache the connections during normal operation, but in tests we want to always start with a clean slate.

es:
> I'm wondering if we could not keep the Providers as instantiated objects, just without passing it the zone object.

IMO the only downside of having instance variables in the Provider is having to clear them during tests. Otherwise I think it makes sense to have the provider classes completely static.

test/zone_test.rb
```diff
@@ -2,6 +2,7 @@
 
 class ZoneTest < Minitest::Test
   def setup
+    @provider = [RecordStore::Provider::DNSimple, RecordStore::Provider::DynECT].sample
```
es: Deleted since it's unused anywhere in the file.

es:
I'll look at creating a pool of provider instances in a follow-up PR.

PR #20, opened 2016-10-28
Fix DNSimple treating SRV attributes as strings

es:
Related: https://github.com/Shopify/record-store/issues/876

r: @klautcomputing @marc-barry


marc-barry:
👏

PR #22, opened 2016-10-31
Cap Zone.modified thread pool size

es:
**Problem**
Before we'd be creating a thread for every zone. This would cause DynECT fetch jobs collide raising errors:
```
"{\"status\": \"failure\", \"data\": {}, \"job_id\": 2950870089, \"msgs\": [{\"INFO\": \"token: This session already has a job running\", \"SOURCE\": \"API-B\", \"ERR_CD\": \"OPERATION_FAILED\", \"LVL\": \"ERROR\"}]}"
```

**Solution**
Use a thread pool for fetching zones with failure handling when fetch jobs collide.

r: @stonith @dwradcliffe @marc-barry

### DIFF
--- a/lib/record_store/zone.rb
+++ b/lib/record_store/zone.rb
@@ -41,11 +41,23 @@ module RecordStore
         end
       end
 
-      def modified
-        modified_zones, mutex = [], Mutex.new
-        self.all.map do |zone|
-          thread = Thread.new do
-            mutex.synchronize {modified_zones << zone} unless zone.unchanged?
+      MAX_PARALLEL_THREADS = 10
+      def modified(verbose: false)
+        modified_zones, mutex, zones = [], Mutex.new, self.all
+
+        (1..MAX_PARALLEL_THREADS).map do
+          Thread.new do
+            current_zone = nil
+            while not zones.empty?
+              mutex.synchronize { current_zone = zones.shift }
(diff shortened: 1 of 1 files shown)


lib/record_store/zone.rb
```diff
@@ -41,11 +41,24 @@ def filter_records(current_records, ignore_patterns)
         end
       end
 
-      def modified
-        modified_zones, mutex = [], Mutex.new
-        self.all.map do |zone|
-          thread = Thread.new do
-            mutex.synchronize {modified_zones << zone} unless zone.unchanged?
+      MAX_PARRALLEL_THREADS = 10
```
etiennebarrie: parallel

lib/record_store/zone.rb
```diff
@@ -41,11 +41,24 @@ def filter_records(current_records, ignore_patterns)
...
+      def modified(verbose: false)
+        modified_zones, mutex, zones = [], Mutex.new, self.all
+
+        (1..MAX_PARRALLEL_THREADS).map do
+          Thread.new do
+            current_zone = nil
+
+            while not zones.empty?
+              mutex.synchronize { current_zone = zones.shift }
```
etiennebarrie: Maybe we should use a `Queue`?

PR #26, opened 2016-11-02
add --name flag to diff command

dwradcliffe:
The code is not the best but it works 😬 
Feel free to rip it up or fix it but I'd love to have this feature.

@es

### DIFF
--- a/lib/record_store/cli.rb
+++ b/lib/record_store/cli.rb
@@ -36,12 +36,15 @@ module RecordStore
       end
     end
 
+    option :name, desc: 'Zone to diff', aliases: '-n', type: :string, required: false
     option :verbose, desc: 'Print records that haven\'t diverged', aliases: '-v', type: :boolean, default: false
     desc 'diff', 'Displays the DNS differences between the zone files in this repo and production'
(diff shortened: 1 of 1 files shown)


lib/record_store/cli.rb
```diff
@@ -36,12 +36,15 @@ def list
...
     end
 
+    option :name, desc: 'Zone to diff', aliases: '-n', type: :string, required: false
     option :verbose, desc: 'Print records that haven\'t diverged', aliases: '-v', type: :boolean, default: false
     desc 'diff', 'Displays the DNS differences between the zone files in this repo and production'
     def diff
-      puts "Diffing #{Zone.defined.count} zones"
+      zones = options.fetch('name').present? ? { options.fetch('name') => Zone.find(options.fetch('name')) } : Zone
+      count = options.fetch('name').present? ? 1 : Zone.defined.count
+      puts "Diffing #{count} zones"
```
es: Instead of count, what about just `zone.keys.size`? You may have to changes `zones` initialization to be ` ? ... : Zone.defined` for this to work though.

PR #28, opened 2017-02-02
Return exit status 1 on failure

etiennebarrie:
When we removed `record-store validate_all_present`, our CI continued to have this as a step and it didn't cause failures.
This made me realize Thor doesn't return a non-zero exit status when the command is unknown.
https://github.com/erikhuda/thor/issues/244#issuecomment-11116989

This probably breaks compatibility in a way, but I think it's fine. Should I ship a 4.1.0 with that?

@es @dwradcliffe


etiennebarrie:
Shipit's not setup, I can't ship a new version. With #29 in, I guess it should be a `v4.1.0`.

PR #29, opened 2017-04-24
Check for circular references in CNAME

vpetkar:
Recently, https://github.com/Shopify/record-store/pull/1271 failed to deploy because of CNAME circular reference.

This PR will validate that this does not happen again.

### DIFF
--- a/lib/record_store/version.rb
+++ b/lib/record_store/version.rb
@@ -1,3 +1,3 @@
 module RecordStore
-  VERSION = '4.0.0'
+  VERSION = '4.0.1'.freeze
 end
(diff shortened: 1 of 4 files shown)


vpetkar:
@es Done!

lib/record_store/version.rb
```diff
@@ -1,3 +1,3 @@
 module RecordStore
-  VERSION = '4.0.0'
+  VERSION = '4.1.0'.freeze
```
es: The version should be bumped to `4.0.1` since we're following [semantic versioning](http://semver.org/) and this is a bug fix 🙂

PR #32, opened 2017-06-27
Move to fog-dynect 0.2.0 and update other gems

charlescng:
* Move fog-dynect to 0.2.0
* Update other gems

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'ejson'
 gem 'fog'
 gem 'fog-json'
 gem 'fog-xml'
-gem 'fog-dynect', '~> 0.1.0'
+gem 'fog-dynect', '~> 0.2.0'
 
 group :test do
   gem 'mocha'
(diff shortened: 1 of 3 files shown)


etiennebarrie:
This is the diff, can the fact that it now includes `ttl` be a problem for us?
https://github.com/fog/fog-dynect/compare/v0.1.0...v0.2.0

etiennebarrie:
It looks like we're using [`get_all_records` directly](https://github.com/Shopify/record_store/blob/master/lib/record_store/provider/dynect.rb#L47), which is the same thing that [`Fog::DNS::Dynect::Records.all`](https://github.com/fog/fog-dynect/blob/v0.2.0/lib/fog/dynect/models/dns/records.rb#L14) uses, but we're just going around it. So we may be able to use that now, but that shouldn't break anything.

charlescng:
I've forgotten about this. A new commit to address:
image

albiere:
We don't seem to use `:ttl => record['ttl']`, which has been added to the 0.2 version. We only use the `rdata` as seen here: https://github.com/Shopify/record_store/blob/master/lib/record_store/provider/dynect.rb#L88

PR #33, opened 2017-11-06
Support double hyphen (--) in CNAME

JamesHageman:
Names such as `website--name.com` are valid domain names according to [this](https://stackoverflow.com/questions/16468309/can-domain-name-have-two-continuous-hyphens) but our CNAME_REGEX doesn't allow them. This PR permits said domain names.


JamesHageman:
@marc-barry I reran the DNSimple and Dyn tests locally with double hyphens in the CNAME and didn't get any failures.

PR #35, opened 2017-11-06
Configure shipit

JamesHageman:


dwradcliffe:
There's a way to do it automatically, I think just name the environment `rubygems` and the file `shipit.rubygems.yml`? See https://github.com/Shopify/statsd-instrument/blob/master/shipit.rubygems.yml for reference

Issue #39, opened 2018-01-05
record_store does not handle edits with multiple entries with the Dynect provider


charlescng:
Looks to have the same underlying problem as https://github.com/Shopify/record_store/issues/27.

Given the scenario where we have two existing TXT entries for the same FQDN:
```
www.example.com TXT foo
www.example.com TXT bar
```
We want to update the `bar` entry to `hello`. 

The existing code does not change the correct entry to `hello` with the Dynect entries. In fact it wants to delete the `foo` entry and update the `bar` entry to `hello`.

I suspect this is also a problem for other record types where there can be multiple entries which as A and AAAA.

The problem does not exist with DNSimple which is supported by Fog::DNS (without the need for another plugin).

charlescng:
The test case isn't quite right. I'll edit it.

charlescng:
The bug is actually somewhere else. I'll have to dig a bit deeper.

sbfaulkner:
re-opened... the bug is (was?) in record_store... and the way it invokes fog-dynect

PR #40, opened 2018-01-05
Add test case for the ambiguous entry edit issue

charlescng:
Add test case for https://github.com/Shopify/record_store/issues/39


charlescng:
OK, this test case doesn't seem quite reproduce the exact problem but it does show the exact symptoms.

PR #41, opened 2018-01-06
fix updates of records with same fqdn

### LINKED ISSUE #39: record_store does not handle edits with multiple entries with the Dynect provider

sbfaulkner:
this PR fixes #39

fog-dynect expects string keys for options hash (https://github.com/fog/fog-dynect/blob/master/lib/fog/dynect/requests/dns/put_record.rb#L19)

record_store was passing symbols (https://github.com/Shopify/record_store/pull/41/files#diff-f3285f7dfe4728480bd7d479256b9608L23)

this fixes that (and includes a test that verifies we actually update a specific record)

### DIFF
--- a/lib/record_store/provider/dynect.rb
+++ b/lib/record_store/provider/dynect.rb
@@ -12,7 +12,7 @@ module RecordStore
       end
 
       def add(record, zone)
-        session.post_record(record.type, zone, record.fqdn, record.rdata, ttl: record.ttl)
+        session.post_record(record.type, zone, record.fqdn, record.rdata, 'ttl' => record.ttl)
       end
 
       def remove(record, zone)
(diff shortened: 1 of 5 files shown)


sbfaulkner:
thanks for shipping this @charlescng

Issue #45, opened 2018-01-23
rake failures


weppos:
On a pristine, newly cloned copy of the repo, I get a mix of warnings and errors when trying to run the tests.

```
➜  record_store git:(master) rake
/Users/weppos/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/activesupport-4.2.9/lib/active_support/core_ext/enumerable.rb:20: warning: method redefined; discarding old sum
/Users/weppos/Code/record_store/lib/record_store/record/a.rb:27: warning: assigned but unused variable - e
/Users/weppos/Code/record_store/lib/record_store/record/aaaa.rb:27: warning: assigned but unused variable - e
/Users/weppos/Code/record_store/lib/record_store/provider.rb:11: warning: assigned but unused variable - e
/Users/weppos/Code/record_store/lib/record_store/provider/dynect.rb:51: warning: assigned but unused variable - e
/Users/weppos/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/fog-core-1.44.3/lib/fog/core/cache.rb:207: warning: assigned but unused variable - name
/Users/weppos/Code/record_store/test/test_helper.rb:29: warning: File.exists? is a deprecated name, use File.exist? instead
/Users/weppos/Code/record_store/test/zone_test.rb:358: warning: ambiguous first argument; put parentheses or a space even after `/' operator
/Users/weppos/Code/record_store/test/zone_test.rb:259: warning: method redefined; discarding old test_zone_unchanged_describes_if_zone_matches_multiple_provider
/Users/weppos/Code/record_store/test/zone_test.rb:249: warning: previous definition of test_zone_unchanged_describes_if_zone_matches_multiple_provider was here
Run options: --seed 49428

# Running:

......../Users/weppos/Code/record_store/lib/record_store/provider.rb:99: warning: File.exists? is a deprecated name, use File.exist? instead
/Users/weppos/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/fog-dynect-0.2.0/lib/fog/dynect/models/dns/record.rb:69: warning: shadowing outer local variable - record
/Users/weppos/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/fog-dynect-0.2.0/lib/fog/dynect/models/dns/zone.rb:33: warning: assigned but unused variable - data
[...]

PR #47, opened 2018-01-23
Switch to fog-dnsimple and API v2

weppos:
`record_store` is currently using the DNSimple adapter from the `fog` gem. This adapter is quite old, it was extracted in 2016 into a `fog-dnsimple`.

Moreover, I've just released a new version of `fog-dnsimple` (fog/fog-dnsimple#3) that uses our API v2. API v1 are deprecated since 2016, and we are now in the process of shutting them down.

I need some help here to update the CircleCI integration to use the new auth mechanism (specifically an API v2 account token you can generate from the Shopify account, and the account ID). /cc @charlescng @es @sbfaulkner 

Technically speaking the client could fetch the ID from the token, but due to the way Fog is structured I decided to request it explicitly.

Unfortunately I was not able to run the integration specs locally because of the fixture failures described in https://github.com/Shopify/record_store/issues/45

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'fog'
 gem 'fog-json'
 gem 'fog-xml'
 gem 'fog-dynect', '~> 0.2.0'
+gem 'fog-dnsimple', '~> 2.0.0'
 gem 'google-cloud-dns'
 
 group :test do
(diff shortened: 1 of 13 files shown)


es:
I fixed the tests and rebased against master.

Original PR LGTM

weppos:
Are you ok with me merging and 🚢  ing? Or do you prefer to deal with it (given your internal infrastructure)?

charlescng:
Merged and tagged. I'll push the gem once CI passes.

weppos:
Awesome, thanks!

PR #48, opened 2018-01-26
Fix runtime dependencies in the gemspec

charlescng:
Add missing gems to the runtime dependencies in the gemspec.

### DIFF
--- a/record_store.gemspec
+++ b/record_store.gemspec
@@ -22,15 +22,19 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'thor'
   spec.add_runtime_dependency 'activesupport', '~> 4.2'
   spec.add_runtime_dependency 'activemodel', '~> 4.2'
-  spec.add_runtime_dependency 'fog'
+  spec.add_runtime_dependency 'ejson'
+
+  spec.add_runtime_dependency 'fog', '>= 1.33.0'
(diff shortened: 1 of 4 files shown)


weppos:
FYI by adding

```ruby
gemspec
```

in the `Gemfile`, you can get rid of the gemspec deps inside the Gemfile. Bundler will load them automatically.

You can probably kill the entire block

https://github.com/Shopify/record_store/blob/7ddd495910c4ed7b6cb18dbaf4047e58c591f9eb/Gemfile#L3-L13

record_store.gemspec
```diff
@@ -22,11 +22,14 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'thor'
   spec.add_runtime_dependency 'activesupport', '~> 4.2'
   spec.add_runtime_dependency 'activemodel', '~> 4.2'
-  spec.add_runtime_dependency 'fog'
+  spec.add_runtime_dependency 'ejson'
+
+  spec.add_runtime_dependency 'fog', '>= 1.33.0'
```
charlescng: Bundler would revert the `fog` gem to 0.7.2 without specifying a version here. I'm not sure why.

charlescng:
@weppos I cleaned up the Gemfile. Thanks.

PR #50, opened 2018-02-12
Loosen active* version constraints

pseudomuto:
# The Problem

This morning I attempted to move an app from tier4 to tier2 using the process outlined here: https://github.com/Shopify/cloudplatform/blob/master/playbooks/cluster-migrations.md

When the deploy attempted to set a timeout on the CloudSQL resource creation, it blew up with:
```
NameError: uninitialized constant ActiveSupport::Duration::ISO8601Parser
```

I tracked this down to _duration_parser.rb_ in the [kubernetes-deploy gem](https://github.com/Shopify/kubernetes-deploy/blob/13f91bfdaa96c6c03b6321a6d81b4b7273a1e4e7/lib/kubernetes-deploy/duration_parser.rb#L19). The issue is that `ActiveSupport::Duration::ISO8601Parser` was introduced in activesupport 5.0.

The [cloudplatform repo](https://github.com/Shopify/cloudplatform) (which contains the migration scripts) has activesupport pinned at `~> 4.2` because that's required by this gem.

As far as I can tell, there's no reason for that to be a requirement (please correct me if I'm wrong).

# The Solution

I've made this gem require activesupport/model `>= 4.2` so that applications (like k8s-deploy or cloudplatform) can use activesupport 5 features.

I also removed the Gemfile.lock, since it doesn't really make sense for gems to include this. [See here](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) for more info.


pseudomuto:
FYI @Shopify/cloudplatform

ibawt:
LGTM, @Edouard-chin any comments here?

record_store.gemspec
```diff
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.0'
 
   spec.add_runtime_dependency 'thor'
-  spec.add_runtime_dependency 'activesupport', '~> 4.2'
-  spec.add_runtime_dependency 'activemodel', '~> 4.2'
```
wvanbergen: I'd be OK with bumping this to 5.1 minimum requirement, but this works too.

charlescng:
We can't deploy this until https://github.com/Shopify/record_store/issues/52 is fixed.

pseudomuto:
Shit, too late: https://github.com/Shopify/record_store/releases/tag/v5.0.2

There's even been another one since

PR #51, opened 2018-02-14
Decrypt secrets on rollback

albiere:
This fixes (maybe) a bug where secrets are not decrypted when a rollback happens while executing `record-store validate_initial_state`.

@Shopify/edgescale

### DIFF
--- a/lib/record_store/cli.rb
+++ b/lib/record_store/cli.rb
@@ -246,12 +246,18 @@ module RecordStore
           `git checkout #{ENV['LAST_DEPLOYED_SHA']}`
           abort "Checkout of old commit failed" if $?.exitstatus != 0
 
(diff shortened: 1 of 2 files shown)


charlescng:
That make sense I guess. Do we need to do it again once it checkout HEAD again? (https://github.com/Shopify/record_store/pull/51/files#diff-2d3e295619bf4546c351b2ac11565563R256)

andrewlouis93:
You'll probably want to bump the version up so you can update it in record-store to get these changes

albiere:
Do we really need to bump the version here? I thought of that, but maybe yes since this is open source project.

charlescng:
The gem won't publish unless you bump the version.

PR #53, opened 2018-02-20
Fix bug where only 30 records are returned

weppos:
API v2 by default paginates the record, and the switch to API v2 caused the call to list records to return only 30 records.

I updated fog-dnsimple to be able to fetch all records. See https://github.com/fog/fog-dnsimple/issues/4

I could not run the specs locally as VCR exploded. FYI the underlying call is now passing pagination parameters, therefore I suspect cassettes has to be updated. Sorry for that. :(

/cc @dturn


sbfaulkner:
@albiere is looking at updating record_store to use ruby-dnsimple directly instead of via fog... if the fixtures/cassettes need updating anyways, that is probably the better path forward

weppos:
I agree. Closing this one for now.

PR #54, opened 2018-02-20
Replace fog/dnsimple with dnsimple gem

albiere:
This PR replaces `fog/dnsimple` with `dnsimple` gem to support version 2 of DNSimple API. Not much changed here besides 3 or 4 calls to the api.

The cassettes have been regenerated again based on a sandbox account created specifically for record_store.

This PR also includes a very minor bugfix related to sensitive data in the cassettes. Both DNSimple and Google add the standard header `Authorization `, however we want to replace the sensitive data with the right placeholder. In the case of this PR we want to replace it with `<DNSIMPLE_API_TOKEN>` and not `<GOOGLE_CLOUD_DNS_AUTH_TOKEN>`.

@Shopify/edgescale


albiere:
Will update the version as soon as we get this reviewed 👍

lib/record_store/provider/dnsimple.rb
```diff
@@ -9,11 +9,12 @@ def supports_alias?
...
           zone,
-          record_hash.fetch(:name),
-          record.type,
-          record_hash.fetch(:content),
+          name: record_hash.fetch(:name),
+          type: record.type,
+          content: record_hash.fetch(:content),
           ttl: record_hash.fetch(:ttl),
           priority: record_hash.fetch(:priority, nil)
         )
```
sbfaulkner: without looking deeper I can't tell whether it would be correct or not, but is this now just..

```
res = session.zones.create_record(
  account_id,
  zone,
  record_hash
)
```

?

albiere: Here's the doc for `#create_record`: http://www.rubydoc.info/gems/dnsimple/Dnsimple/Client/ZonesRecords#create_record-instance_method

I guess that will do it.

lib/record_store/provider/dnsimple.rb
```diff
@@ -29,42 +30,50 @@ def add(record, zone)
...
-          provider: 'DNSimple',
-          dnsimple_token: secrets.fetch('api_token'),
-          dnsimple_account: secrets.fetch('account_id'),
-        }
+      def account_id
+        @account_id ||= secrets.fetch('account_id')
+      end
+
+      def parameterize_record(record)
+        record.instance_variables.each_with_object({}) do |attr, hash|
```
sbfaulkner: this smells a little to me... what type is `record`?

lib/record_store/provider/dnsimple.rb
sbfaulkner: mostly I didn't know what would wind up in the record_hash... ie. is `type` included or not? are there additional attributes, or just these ones... 🤷‍♂️

lib/record_store/provider/dnsimple.rb
sbfaulkner: ie. what class?

lib/record_store/provider/dnsimple.rb
albiere: Yes. `type` is included in the `record_hash` right here: https://github.com/Shopify/record_store/pull/54/files#diff-edcaaf931e2c87e8e59ba598100e8095R128

lib/record_store/provider/dnsimple.rb
albiere: This is just a translation from a object to a hash to avoid changing other methods. Here's the type of the object: http://www.rubydoc.info/gems/dnsimple/Dnsimple/Struct/ZoneRecord

albiere: Just removed the translation layer and update the `build_from_api` to accommodate dnsimple objects.

lib/record_store/provider/dnsimple.rb
```diff
@@ -29,76 +22,75 @@ def add(record, zone)
...
         when 'SPF', 'TXT'
-          record.merge!(txtdata: api_record.fetch('content').gsub(';', '\;'))
+          record.merge!(txtdata: api_record.content.gsub(';', '\;'))
         when 'SRV'
-          weight, port, host = api_record.fetch('content').split(' ')
+          weight, port, host = api_record.content.split(' ')
 
           record.merge!(
-            priority: api_record.fetch('priority').to_i,
+            priority: api_record.priority.to_i,
```
weppos: FYI if this is coming from the API, `priority` should be an integer. We normalized it to be always an integer. But it doesn't hurt to keep `to_i`, I'm not sure if it was in response of some specific issue in the past.

albiere: Cool! I tried to make as less changes as possible. Will update accordingly now. Thanks for pointing it out. 👍

es:
This is awesome 😻

PR #55, opened 2018-03-31
Refactor for more consistent output

sbfaulkner:
I noticed that when the GoogleCloudDNS provider was used, that some of the progress messages were not produced.

The ones for specific operations (Removing/Creating/Updating) are not as simple, but this PR at least moves the Apply and Published output to the common function in the changeset that invokes the individual providers.

context: for dynect and dnsimple we have to actually apply the individual changes (which is what the base implementation of `apply_changeset` does... however, google never calls the base implementation and instead overrides it

### DIFF
--- a/lib/record_store/changeset.rb
+++ b/lib/record_store/changeset.rb
@@ -57,7 +57,11 @@ module RecordStore
     end
 
     def apply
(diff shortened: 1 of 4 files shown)


albiere:
> The ones for specific operations (Removing/Creating/Updating) are not as simple
Can you elaborate a little bit more about it?

What's the real benefit of moving those two logs to `Changeset#apply`?

PR #62, opened 2018-06-21
Use distinct Dynect sessions for each thread

sbfaulkner:
fog-dynect is not expecting to run with multiple threads sharing the dynect session, but that’s how record_store is using it

* fog-dynect issues a request and then if the response is  a 307 it polls the jobs api until the request is completed
* but… meanwhile another thread could have attempted to issue a request with the api

which means the code here... https://github.com/Shopify/record_store/blob/master/lib/record_store/zone.rb#L56
could cause us to thrash on the api

this PR attempts to address that by adding a mutex and syncronizing for exclusive session use

### DIFF
--- a/lib/record_store/provider/dynect.rb
+++ b/lib/record_store/provider/dynect.rb
@@ -4,15 +4,15 @@ module RecordStore
   class Provider::DynECT < Provider
     class << self
       def freeze_zone(zone)
-        synchronize { session.put_zone(zone, freeze: true) }
+        session.put_zone(zone, freeze: true)
       end
 
       def thaw_zone(zone)
-        synchronize { session.put_zone(zone, thaw: true) }
+        session.put_zone(zone, thaw: true)
       end
 
       def publish(zone)
-        synchronize { session.put_zone(zone, publish: true) }
+        session.put_zone(zone, publish: true)
       end
 
       # Applies changeset to provider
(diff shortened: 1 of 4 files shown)


sbfaulkner:
ADDITIONAL NOTE: we remove the "session collision detection" since this is actually a really bad thing... ie. we should NEVER have an outstanding job when making a request now since we fog-dynect polls for completion and the critical section should prevent it

lib/record_store/provider/dynect.rb
```diff
@@ -105,6 +106,11 @@ def build_from_api(api_record)
 
         Record.const_get(type).new(record)
       end
+
+      def synchronize(&block)
+        @mutex ||= Mutex.new
```
charlescng: Can we not define this as an instance of the class instance instead of doing the lazy loading that has the theoretical race condition?

sbfaulkner:
> Do we have to use the same session? Are we allowed to use a session per thread?

The isssue here (I think) would be rate limit but not sure of the history here.

lib/record_store/provider/dynect.rb
sbfaulkner: there is no instance :-( it's all class methods :-(

sbfaulkner: I could assign it at load time though :-/

sbfaulkner: eg. 
```
      @mutex = Mutex.new

      def synchronize(&block)
        @mutex.synchronize(&block)
      end
```

charlescng: 👍 That's what I mean.

charlescng:
I guess we can try it as-is and see what the impact is.

sbfaulkner:
@charlescng trying the simpler approach you suggested

sbfaulkner:
@charlescng I dismissed your review since I changed to try your approach instead

PR #66, opened 2018-07-05
Add support for underscores in hostname CNAME entries

JeremyAshton-shopify:
AWS Certificate Manager leverages DNS CNAME entries to verify ownership of domains for certificate issuance as noted at https://docs.aws.amazon.com/acm/latest/userguide/gs-acm-validate-dns.html.  Although technically allowed, it is rare to see these in the hostname portion of CNAME entries.

@sbfaulkner @charlescng


sbfaulkner:
you'll want to bump the version to `5.1.1` in https://github.com/Shopify/record_store/blob/master/lib/record_store/version.rb too

PR #67, opened 2018-07-06
limit dynect api request rate

sbfaulkner:
use https://github.com/Shopify/limiter to restrict the dynect request rate

given that the gem has tests of its own... I'm not convinced there's a need for additional tests here (since existing tests work)

since I'm adding a GEM dependency, I think I may be more interested in getting y'all to review it before merging this... see: https://github.com/Shopify/limiter

wdyt @albiere ?

### DIFF
--- a/lib/record_store/provider/dynect.rb
+++ b/lib/record_store/provider/dynect.rb
@@ -1,4 +1,8 @@
 require 'fog/dynect'
+require 'limiter'
+
+Fog::DNS::Dynect::Real.extend Limiter::Mixin
+Fog::DNS::Dynect::Real.limit_method :request, rate: 5, interval: 1 # 5 RPS == 300 RPM
 
 module RecordStore
   class Provider::DynECT < Provider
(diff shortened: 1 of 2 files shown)


sbfaulkner:
@charlescng should we compare run time with throttles of...

5 / second
300 / minute
150 / 30 seconds
75 / 15 seconds

or should we ship this, and then try again with one of the other rates if it's a LOT slower

charlescng:
Let's ship this as is. It might not be a problem and revisit if it's super slow.

sbfaulkner:
fwiw... I originally made the 5 rps decision to smooth out the rate of requests (instead of allowing a burst of 300 requests, and then having a 1 minute delay before it could continue) - but yes, I think observing impact is the simple path for now

charlescng:
Another thought is that the rate _should_ be configurable. I assume that different accounts will have different rate limits.

sbfaulkner:
> Another thought is that the rate should be configurable. I assume that different accounts will have different rate limits.

yeah, that would be a good idea - opened issue #68

sbfaulkner:
bumped version and updated changelog

cttttt:
Had a late look and 👍

PR #70, opened 2018-10-12
Add regexp based ignores

cttttt:
**What's Changed? Why?**

Presently, a RecordStore zone can be configured to ignore records that fit a list of static `ignore_patterns`.  Records in zones are compared against these static patterns and any records with key/value pairs that are an **exact match** for an ignore pattern's will be ignored.

This change introduces a `match` field in `ignore_patterns` that controls a pattern's matching criteria: either `exact` or `regex`.  This gives maintainers of zone data more flexibility to define classes of records that will be ignored.

**Example**

For example, given the following `record_store` zone configuration:

```
one-record.com:
  config:
    providers:
    - DynECT
    ignore_patterns:
    - type: NS
      fqdn: one-record.com.
      match: exact
    - type: TXT
      fqdn: "^_acme-challenge\."
      match: regex
    - type: NS
      fqdn: b.one-record.com.
  records: 
    - type: A
      fqdn: a.one-record.com.
      ttl: 3600
      address: 10.10.10.10
```
...records like the following would be ignored by virtue of the `regex` `ignore_pattern`:

```
- type: TXT
  fqdn: _acme_challenge.a.one-record.com.
  ttl: 3600
  txtdata: AFKJDSJFLKSJHDJ
```

...and records like the following would be ignored by virtue of the `exact` `ignore_pattern`:

```
- type: NS
  fqdn: one-record.com.
  ttl: 3600
  nsdname: ns1.p19.dynect.net.
```

...and records like the following would be ignored by virtue of the implicitly `exact` `ignore_pattern`:

```
- type: NS
  fqdn: b.one-record.com.
  ttl: 3600
  nsdname: ns1.p19.dynect.net.
```

...while the record in the zone file will not be ignored.

**Notes**

- Altered the VCR redaction config to support a newer version of `googleauth` gem.


albiere:
@cttttt I'm not sure I like the interface for the client. Even though I understand that works. 

`ignore_patterns` and `ignore_regexes` are too similar to each other. Is there a way we could detect the ignore pattern is a regex or a "full string"? Maybe we could detect a regex when the "value" is specified as `/<some string>/`.

If that's not possible, what do you think about giving the record a type. For example

```
ignore_patterns:
    - match: (string|regex) 
       type: NS
       fqdn: one-record.com.
```

Given we also need to cover existing zone files, we could define a default valur for the key `match`. What do you think?

cttttt:
(Thinking out loud)

- Trying to detect a regex feels safe for most types of records (where it doesn't make sense to have a value with a leading non-alpha character).  But for TXT records, we could break users, as a text record can contain pretty much anything.
- As we chatted a little about, the schema of an `ignore_pattern` is super crisp: The fields in each object corresponding to the fields in a zone record.  If a zone record contains all of the key/value pairs from an `ignore_pattern`, we should ignore the zone record.
  - Reserving the field name, `match` muddies this schema of `ignore_pattern`s with a reserved field name, `match`.
  - But it does neaten up a zone config some having just `ignore_pattern`'s...and even I was surprised that `ignore_pattern`'s weren't even patterns.

(/Thinking out load)

I'll give it a shot.  Mind if I change `string` to `exact` as that default value for `match`?  Feels like a more clear description of how the matching will be performed.

cttttt:
@albiere please have another look

lib/record_store/zone/config/ignore_pattern.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        def value_matches?(ignore_pattern_value, record_value)
+          return record_value.match?(ignore_pattern_value) if is_regex?
+          return record_value === ignore_pattern_value if is_exact?
+          false
+        end
+
+        def is_regex?
+          orig_hash['match'] === 'regex'
```
albiere: Any real special reason on using "===" rather than "==". Isn't that too defensive?

lib/record_store/zone/config/ignore_pattern.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        def is_exact?
+          !orig_hash.key?('match') || (orig_hash['match'] === 'exact')
```
albiere: `origin_hash['match'] == "exact"` is more than enough here, right? Ruby will return `nil` if you there's no key named `match` and that will evaluate the condition to false.

lib/record_store/zone/config/ignore_pattern.rb
albiere: Nitpick... we could use constants for the two strings `regex` and `exact`.

lib/record_store/zone.rb
```diff
@@ -32,10 +32,10 @@ def download(name, provider_name, **write_options)
         zone.write(**write_options)
       end
 
-      def filter_records(current_records, ignore_patterns)
-        ignore_patterns.inject(current_records) do |remaining_records, pattern|
+      def filter_records(current_records, ignore_objects)
+        ignore_objects.inject(current_records) do |remaining_records, pattern|
           remaining_records.reject do |record|
-            pattern.all? { |(key, value)| record.respond_to?(key) && value === record.send(key) }
+            pattern.should_ignore?(record)
```
albiere: Nitpick... since the calls are pretty short, we could just make this a one-liner like so:

```ruby
remaining_records.reject { |record| pattern.should_ignore?(record) }
```

lib/record_store/zone/config.rb
```diff
@@ -24,7 +27,7 @@ def supports_alias?
       def to_hash
         config_hash = {
           providers: providers,
-          ignore_patterns: ignore_patterns,
+          ignore_patterns: ignore_patterns.map {|ignore_pattern| ignore_pattern.to_hash },
```
albiere: You can use `ignore_patterns.map(&:to_hash)` and that will have the same result.

lib/record_store/zone/config.rb
```diff
@@ -8,7 +8,10 @@ class Config
       validate :validate_zone_config
 
       def initialize(ignore_patterns: [], providers: nil, supports_alias: nil)
-        @ignore_patterns = ignore_patterns
+        @ignore_patterns = ignore_patterns.map do|ignore_pattern|
```
albiere: There's a space missing between `do` and `|`. Sorry :-/

lib/record_store/zone/config/ignore_pattern.rb
```diff
@@ -0,0 +1,36 @@
...
+          return record_value.match?(ignore_pattern_value) if is_regex?
+          return record_value === ignore_pattern_value if is_exact?
+          false
+        end
+
+        def is_regex?
+          orig_hash['match'] === 'regex'
+        end
+
+        def is_exact?
```
albiere: In ruby we usually don't use `is_<attr>` or `set_attr` or `get_attr`. Since ruby allows you to use special symbols as part of the method name, you can have `<attr>?`, `<attr>=` and `<attr>` as the equivalent version in Ruby.

lib/record_store/zone/config.rb
albiere: I've seen this pattern in may different places in your code. Maybe there's something in your editor. Maybe that's good 🤷‍♂️  and I'm just not used to that.

lib/record_store/zone/config/ignore_pattern.rb
```diff
@@ -0,0 +1,41 @@
...
+          all_pairs.all?{|(key, value)| record.respond_to?(key) && value_matches?(value, record.send(key))}
+        end
+
+        def to_hash
+          orig_hash
+        end
+
+        private
+
+        attr_accessor :orig_hash
```
albiere: Do you need this to be private? If you don't want to expose the `orig_hash` publicly, you should use the instance reference with `@orig_hash` in your code.

lib/record_store/zone/config.rb
cttttt: Will update it if it devistes from the rest of the code.  If there's a linter config, I'll update it to enforce this.

sbfaulkner:
maybe update description for the updated schema for this?

albiere:
@cttttt Let's also update the README to include this new feature

lib/record_store/zone/config/ignore_pattern.rb
cttttt: My train of thought here:

- I created an `attr_accessor` for `orig_hash`
- I created a public accessor, `to_hash` for serializing
- Then I was like _orig_hash should be private_ and did this

I only need `to_hash`.  Imma dumb.

albiere: Don't worry man!

cttttt:
I'll see about tackling the README in a new PR.

@sbfaulkner , @albiere 👀 ?

albiere:
@cttttt sounds good 👍

lib/record_store/zone.rb
```diff
@@ -32,10 +32,10 @@ def download(name, provider_name, **write_options)
         zone.write(**write_options)
       end
 
-      def filter_records(current_records, ignore_patterns)
-        ignore_patterns.inject(current_records) do |remaining_records, pattern|
+      def filter_records(current_records, ignore_objects)
+        ignore_objects.inject(current_records) do |remaining_records, pattern|
```
sbfaulkner: What’s in a name? If the collection is objects instead of patterns, why are the individual items being enumerated still `pattern`?

lib/record_store/zone/config/ignore_pattern.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          false
+        end
+
+        def regex?
+          to_hash[MATCH_TYPE_FIELD] == MATCH_TYPE_REGEX
+        end
+
+        def exact?
+          !to_hash.key?(MATCH_TYPE_FIELD) || \
```
sbfaulkner: Hash#fetch supports passing a default ... to_hash.fetch(MATCH_TYPE_FIELD, MATCH_TYPE_EXACT)

lib/record_store/zone/config/ignore_pattern.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+
+        private
+
+        def all_pairs
+          to_hash.reject { |key| key == MATCH_TYPE_FIELD }
+        end
+
+        def value_matches?(ignore_pattern_value, record_value)
+          return record_value.match?(ignore_pattern_value) if regex?
```
sbfaulkner: The conditions feel like an OOP antipattern. You were creating two subclasses originally right?

lib/record_store/zone.rb
cttttt: This is a holdover from a previous iteration where regexp based ignore patterns had a different type. I later took Angelo's advice and just added a field, so this change is unnecessary. Will revert this hunk.

PR #71, opened 2018-10-22
Support dnsimple-4.5.0

cttttt:
- CI broke:
  - This PR will be in the record-store repo, so I should be able to trigger CI w/o issue and verify before merging (doing so manually because Github has turned off webhooks temporarily)
  - Seems like:
  `record_store --<depends on>--> dnsimple-ruby --<depends on>-- dnsimple >=2.0`
    I tested locally with a setup that had the   DNSimple 4.4.0 gem and it worked fine.  However, the DNSimple 4.5.0 gem changed a few method names, and is causing trouble in CI.

Will stop messing around with fork repos and create a branch here, and ensure that passes CI before merging.

Will try to get ... some gemspec dependency tightened up as a follow up.

### DIFF
--- a/lib/record_store/provider/dnsimple.rb
+++ b/lib/record_store/provider/dnsimple.rb
@@ -9,7 +9,7 @@ module RecordStore
 
       # returns an array of Record objects that match the records which exist in the provider
       def retrieve_current_records(zone:, stdout: $stdout)
-        session.zones.all_records(account_id, zone).data.map do |record|
+        session.zones.all_zone_records(account_id, zone).data.map do |record|
           begin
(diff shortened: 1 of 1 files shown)


weppos:
Sorry for the extra effort. We packaged a release that included a compatibility break my mistake.

cttttt:
@weppos 'Saul Goodman

cttttt:
Will merge when we're ready to move up.  To make progress more quickly, we just switched to depending on the older version of the dnsimple gem directly, instead of including it via the deprecated (renamed) dnsimple-ruby.

XaF:
😭

PR #72, opened 2018-10-22
Just use dnsimple directly

cttttt:
dnsimple-ruby was renamed to dnsimple; this gives us an opportunity to depend on a specific version of dnsimple and workaround our issues on master.


record_store.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'fog-json'
   spec.add_runtime_dependency 'fog-xml'
   spec.add_runtime_dependency 'fog-dynect', '~> 0.2.0'
-  spec.add_runtime_dependency 'dnsimple-ruby', '~> 2.0'
```
weppos: 👍

PR #74, opened 2019-01-30
fix tests for newer thor

sbfaulkner:
this PR fixes tests to work with newer Thor gem

older versions were passing the wrong type to `Kernel#exit`

### DIFF
--- a/record_store.gemspec
+++ b/record_store.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.0'
 
-  spec.add_runtime_dependency 'thor'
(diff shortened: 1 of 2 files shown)


record_store.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.0'
 
-  spec.add_runtime_dependency 'thor'
+  spec.add_runtime_dependency 'thor', '~> 0.20.3', '>= 0.20.3'
```
charlescng: Isn't just `~> 0.20.3` enough?

sbfaulkner: updated

PR #75, opened 2019-01-31
lowercase fqdns

sbfaulkner:
BACKGROUND:
* DNS is case-insenstive
* record-store is (was?) case sensitive
* DNSimple does not allow creation of DNS records with Mixed-Case fqdns
* DNSimple did support it at some point in history (we have some existing ones)
* Dynect supports creation of DNS records with Mixed-Case fqdns
* Dynect creates DNS records with "sticky" case - ie. once created in Mixed-Case it will always be mixed-case, even if DELETED, PUBLISHED and RE-CREATED

This set of conditions lead to being able to create a zone definition which attempts to create a mixed-case DNS record, but that immediately diverges state (or is considered to have diverged)

this PR downcases the fqdn...

1. when loading from API
2. when loading from YAML

the end result will be to...

1. always create DNS records with lower-case fqdns
2. treat all existing DNS records as if they were lower-case

### DIFF
--- a/lib/record_store/provider/dnsimple.rb
+++ b/lib/record_store/provider/dnsimple.rb
@@ -97,9 +97,8 @@ module RecordStore
           )
         end
 
-        unless record.fetch(:fqdn).ends_with?('.')
-          record[:fqdn] += '.'
-        end
+        fqdn = record.fetch(:fqdn)
+        record[:fqdn] = "#{fqdn}." unless fqdn.ends_with?('.')
 
         Record.const_get(record_type).new(record)
       end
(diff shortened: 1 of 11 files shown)


lib/record_store/provider/dnsimple.rb
```diff
@@ -64,11 +64,14 @@ def secrets
       end
 
       def build_from_api(api_record, zone)
+        fqdn = api_record.name.present? ? "#{api_record.name}.#{zone}" : zone
+        fqdn = "#{fqdn}." unless fqdn.ends_with?('.')
```
cttttt: Seems like all Providers need to add a dot suffix to a string.  Consider promoting this operation to Provider.

sbfaulkner: I don’t disagree but also out of scope - so not right now. :-)

PR #77, opened 2019-03-13
add CAA record support

sbfaulkner:
so, you want to add CAA records to a zone... easy, right?

nope. record_store doesn't support CAA records. doh!

so, this PR adds that support.

DNSimple returns an rdata with of form `<flags> <tag> "<value>"`
Dyn returns component rdata fields of `flags`, `tag` and `value`
Google returns a record set as an array of rdata which are formatted similar to DNSimple


lib/record_store/record/caa.rb
```diff
@@ -0,0 +1,46 @@
+module RecordStore
+  class Record::CAA < Record
+    attr_accessor :flags, :tag, :value
+
+    LABEL_REGEX = '[a-z0-9](?:-*[a-z0-9])*'
+    DOMAIN_REGEX = /\A#{LABEL_REGEX}(?:\.#{LABEL_REGEX})\z/
```
albiere: This regex is not good enough.... something like `domain.c` would match. You need to consider at least 2 characters for the TLD.

charlescng: How does this differ from FQDN from the parent class? I guess this is needed to handle `mailto:` values.

sbfaulkner: I've followed the [RFC](https://tools.ietf.org/html/rfc6844#section-5.2) for this...

```
   domain = label *("." label)
   label = (ALPHA / DIGIT) *( *("-") (ALPHA / DIGIT))
```

that said, I meant to make this a case-insensitive REGEX

sbfaulkner: see my reply to angelo's review re: RFC and domain value... the URLs are not handled with this regex and are only validated for `iodef` tags

albiere: I thought the minimum size of a TLD was 2, but haven't really found anything to support that so 👍

sbfaulkner: I believe you're correct... but I'm basing this on the RFC for the record rather than the commonly accepted idea of a domain

PR #81, opened 2019-08-21
Long txt fields

mredan:
TXT records should be limited to 255 chars. 

if you want longer ones, just [quote](https://kb.isc.org/docs/aa-00356) your way around it.

This will automatically add quotes to TXT records that are longer than 255 chars.


lib/record_store/record.rb
```diff
@@ -17,7 +17,8 @@ def escape(value)
       end
 
       def quote(value)
-        %("#{escape(value)}")
+        Record.needs_long_quotes?(value) ? result = escape(value).scan(/.{1,255}/).join('" "') : result = escape(value)
```
charlescng: The assignment can be done outside the ternary operator
```ruby
result = Record.needs_long_quotes?(value) ? escape(value).scan(/.{1,255}/).join('" "') : escape(value)
```

mredan:
tested against prod
```
$ ../record_store/bin/record-store diff
SPF record type is deprecated (See RFC 7208 Section 14.1)
SPF record type is deprecated (See RFC 7208 Section 14.1)
Diffing 179 zones
.....................................................................
Zone: metabunny.biz
--------------------
Provider: DNSimple
Add:
 - [TXTRecord] random.metabunny.biz. 3600 IN TXT "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
--------------------
Provider: DynECT
Add:
 - [TXTRecord] random.metabunny.biz. 3600 IN TXT "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
====================
............................................SPF record type is deprecated (See RFC 7208 Section 14.1)
............................SPF record type is deprecated (See RFC 7208 Section 14.1)
SPF record type is deprecated (See RFC 7208 Section 14.1)
.....................................
```

the `SPF record type is deprecated` are unrelated.

lib/record_store/record.rb
```diff
@@ -17,7 +17,8 @@ def escape(value)
       end
 
       def quote(value)
-        %("#{escape(value)}")
+        result = Record.needs_long_quotes?(value) ? escape(value).scan(/.{1,255}/).join('" "') : escape(value)
```
XaF: Can't we use directly `needs_long_quotes?` without specifying the class given we're already in the same place?

lib/record_store/record/txt.rb
```diff
@@ -2,28 +2,37 @@ module RecordStore
...
-    validates :txtdata, presence: true, length: { maximum: 255 }
-    validate :escaped_semicolons
+    validates :txtdata, presence: true, length: { maximum: 4096 }
+    validate :escaped_semicolons, :long_txt_quoting
 
     def initialize(record)
       super
-      @txtdata = record.fetch(:txtdata)
+      txt = record.fetch(:txtdata)
+      Record.needs_long_quotes?(txt) ? @txtdata = Record.quote(txt) : @txtdata = txt
```
XaF: Same thing as Chuck said for the assignment. Rubocop will complain about this.
Same question for the repeats of `Record.` as above

lib/record_store/record/txt.rb
```diff
@@ -2,28 +2,37 @@ module RecordStore
...
     end
 
     def rdata
       { txtdata: txtdata }
     end
 
     def rdata_txt
-      Record.quote(txtdata)
+      txtdata.length > 255 ? txt = txtdata : txt = Record.quote(txtdata)
```
XaF: Same two as above.
And as you're returning `txt` anyway, assignments are redundant here

lib/record_store/record/txt.rb
```diff
@@ -2,28 +2,37 @@ module RecordStore
...
+      txt
     end
 
     private
 
+    def long_txt_quoting
+      if Record.needs_long_quotes?(txtdata)
```
XaF: Same for `Record`

lib/record_store/record/txt.rb
```diff
@@ -2,28 +2,37 @@ module RecordStore
   class Record::TXT < Record
     attr_accessor :txtdata
 
-    validates :txtdata, presence: true, length: { maximum: 255 }
-    validate :escaped_semicolons
+    validates :txtdata, presence: true, length: { maximum: 4096 }
```
charlescng: Is this an arbitrary limit? I took a cursory look at [RFC 4408](https://datatracker.ietf.org/doc/rfc4408/?include_text=1) and I didn't see any obvious length limit.

XaF: From your link:

>   SPF or TXT records containing multiple strings are useful in
   constructing records that would exceed the 255-byte maximum length of
   a string within a single TXT or SPF RR record

mredan:
tested on metabunny.biz, and it failed with DNSimple.

the record was pushed successfully, but it could not be queried (NXDOMAIN returned). I could see it in the DNSimple WebUI though.


```
$ dig +short TXT random2.metabunny.biz @ns1.p19.dynect.net.
"bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"

$ dig +short TXT random2.metabunny.biz @ns1.dnsimple.com.
$
```


no mention of this in  https://support.dnsimple.com/articles/txt-record/.
But from https://support.dnsimple.com/articles/dkim-record/ it says to not include quotes.

lib/record_store/record/txt.rb
charlescng: Yes, I was referring to the 4096 byte limit. :)

XaF: Huh my bad. Should have read what line you commented on!

mredan: I can't find the reference any more, but I had read that there was a general limit of 4096 bytes on any DNS record.

..but then, I see that [RDLENGTH](https://www.rfc-editor.org/rfc/rfc1035.html#section-3.2.1) is a 16-bit integer..so really, the theoretical max size of a record is 65535 bytes.

I'll look this up again, and adjust as needed.

charlescng: I'm OK with a 4096 byte limit. In practice, this won't be common and we can always bump the limit if needed.

mredan:
DNSimple requires long TXT records to be stored as just one long string..they manipulate it to be broken into 255-char quoted strongs.

Dyn does not allow this, nor does [ns1](https://help.ns1.com/hc/en-us/articles/360022091074-Understanding-DNS-record-types) nor [google](https://cloud.google.com/dns/records/).

so we convert long txt records to the quoted version, then undo that when talking to DNSimple.

mredan:
@XaF @charlescng can you guys 👀 again pls? 
it seems to work now at least :)

charlescng:
> so we convert long txt records to the quoted version, then undo that when talking to DNSimple.

Does it make sense to push the quoting and splitting logic down to the provider instead so we don't have to do that?

mredan:
> Does it make sense to push the quoting and splitting logic down to the provider instead so we don't have to do that?

I updated it to push splitting logic down to the providers, and tested @charlescng

lib/record_store/provider/dynect.rb
```diff
@@ -89,7 +89,7 @@ def secrets
       def api_rdata(record)
         case record.type
         when 'TXT'
```
charlescng: Add SPF here too?

charlescng:
Can you make sure you rebase on origin/master and bump the version file?

charlescng:
@mredan We forgot to update the CHANGELOG: https://github.com/Shopify/record_store/blob/master/CHANGELOG.md

mredan:
> @mredan We forgot to update the CHANGELOG:
I'm a bad person.

updating now.

PR #82, opened 2019-09-18
Add NS1 provider

gooallen:
**What's the problem?**

NS1 is not a supported DNS provider

**How will this fix it?**

- By adding a provider for NS1

**Notes**

- To record new cassettes you'll need:
  - An NS1 account and API key
  - The empty zone, `test.recordstore.io` in your account.

~~**Why WIP?**~~

- ~~[x] We should refine how we handle errors from the NS1 library``~~
- ~~[x] We should improve how we remove records.  Right now, removing a record will remove all records for the FQDN.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dev/
 pkg/
 Gemfile.lock
 .bundle/config
+.byebug_history
--- a/lib/record_store.rb
+++ b/lib/record_store.rb
@@ -31,6 +31,7 @@ require 'record_store/provider'
 require 'record_store/provider/dynect'
 require 'record_store/provider/dnsimple'
 require 'record_store/provider/google_cloud_dns'
+require 'record_store/provider/ns1'
 require 'record_store/cli'
 
 module RecordStore
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -19,6 +19,8 @@ module RecordStore
           'DynECT'
         when /googledomains\.com\z/
           'GoogleCloudDNS'
+        when /\.nsone\.net\z/
+          'NS1'
         else
           nil
         end
--- a/lib/record_store/provider/ns1.rb
+++ b/lib/record_store/provider/ns1.rb
@@ -1,0 +1,220 @@
+require_relative 'ns1/client'
+
+module RecordStore
+  class Provider::NS1 < Provider
+    class Error < StandardError; end
+
+    class << self
+      def client
+        Provider::NS1::Client.new(api_key: secrets['api_key'])
+      end
+
+      # Downloads all the records from the provider.
+      #
+      # Returns: an array of `Record` for each record in the provider's zone
+      def retrieve_current_records(zone:, stdout: $stdout)
+        full_api_records = records_for_zone(zone).map do |short_record|
+          client.record(
+            zone: zone,
+            fqdn: short_record["domain"],
+            type: short_record["type"],
+            must_exist: true,
+          )
+        end
+
+        full_api_records.map { |r| build_from_api(r, zone) }.flatten.compact
+      end
+
+      # Returns an array of the zones managed by provider as strings
+      def zones
+        client.zones.map { |zone| zone['zone'] }
+      end
+
+      private
+
+      # Fetches simplified records for the provided zone
+      def records_for_zone(zone)
+        client.zone(zone)["records"]
+      end
+
+      # Creates a new record to the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # record - a kind of `Record`
+      def add(record, zone)
+        new_answers = [{ answer: build_api_answer_from_record(record) }]
+
+        record_fqdn = record.fqdn.sub(/\.$/, '')
+
+        existing_record = client.record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type
+        )
+
+        if existing_record.nil?
+          client.create_record(
+            zone: zone,
+            fqdn: record_fqdn,
+            type: record.type,
+            params: { answers: new_answers, ttl: record.ttl }
+          )
+          return
+        end
+
+        existing_answers = existing_record['answers'].map { |answer| symbolize_keys(answer) }
+        client.modify_record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          params: { answers: existing_answers + new_answers, ttl: record.ttl }
+        )
+      end
+
+      # Deletes an existing record from the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # record - a kind of `Record`
+      def remove(record, zone)
+        record_fqdn = record.fqdn.sub(/\.$/, '')
+
+        existing_record = client.record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type
+        )
+        return if existing_record.nil?
+
+        pruned_answers = existing_record['answers']
+          .map { |answer| symbolize_keys(answer) }
+          .reject { |answer| answer[:answer] == build_api_answer_from_record(record) }
+
+        if pruned_answers.empty?
+          client.delete_record(
+            zone: zone,
+            fqdn: record_fqdn,
+            type: record.type
+          )
+          return
+        end
+
+        client.modify_record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          params: { answers: pruned_answers }
+        )
+      end
+
+      # Updates an existing record in the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # id - provider specific ID of record to update
+      # record - a kind of `Record` which the record with `id` should be updated to
+      def update(id, record, zone)
+        record_fqdn = record.fqdn.sub(/\.$/, '')
+
+        existing_record = client.record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          must_exist: true,
+        )
+
+        # Identify the answer in this record with the matching ID, and update it
+        updated = false
+        existing_record["answers"].each do |answer|
+          next if answer["id"] != id
+          updated = true
+          answer["answer"] = build_api_answer_from_record(record)
+        end
+
+        raise(Error, "while trying to update a record, could not find answer with fqdn: #{record.fqdn}, type; #{record.type}, id: #{id}") unless updated
+
+        client.modify_record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          params: { answers: existing_record["answers"], ttl: record.ttl }
+        )
+      end
+
+      def build_from_api(api_record, zone)
+        fqdn = Record.ensure_ends_with_dot(api_record["domain"])
+
+        record_type = api_record["type"]
+        return if record_type == 'SOA'
+
+        api_record["answers"].map do |api_answer|
+          answer = api_answer["answer"]
+          record = {
+            ttl: api_record["ttl"],
+            fqdn: fqdn.downcase,
+            record_id: api_answer["id"]
+          }
+
+          case record_type
+          when 'A', 'AAAA'
+            record.merge!(address: answer.first)
+          when 'ALIAS'
+            record.merge!(alias: answer.first)
+          when 'CAA'
+            flags, tag, value = answer
+
+            record.merge!(
+              flags: flags.to_i,
+              tag: tag,
+              value: Record.unquote(value),
+            )
+          when 'CNAME'
+            record.merge!(cname: answer.first)
+          when 'MX'
+
+            preference, exchange = answer
+
+            record.merge!(
+              preference: preference.to_i,
+              exchange: exchange,
+            )
+          when 'NS'
+            record.merge!(nsdname: answer.first)
+          when 'SPF', 'TXT'
+            record.merge!(txtdata: Record.unescape(answer.first).gsub(';', '\;'))
+          when 'SRV'
+            priority, weight, port, host = answer
+
+            record.merge!(
+              priority: priority.to_i,
+              weight: weight.to_i,
+              port: port.to_i,
+              target: Record.ensure_ends_with_dot(host),
+            )
+          end
+          Record.const_get(record_type).new(record)
+        end
+      end
+
+      def build_api_answer_from_record(record)
+        if record.is_a?(Record::MX)
+          [record.preference, record.exchange]
+        elsif record.is_a?(Record::TXT) or record.is_a?(Record::SPF)
+          [record.txtdata]
+        elsif record.is_a?(Record::CAA)
+          [record.flags, record.tag, record.value]
+        elsif record.is_a?(Record::SRV)
+          [record.priority, record.weight, record.port, record.target]
+        else
+          [record.rdata_txt]
+        end
+      end
+
+      def symbolize_keys(hash)
+        hash.map{ |key, value| [key.to_sym, value] }.to_h
+      end
+
+      def secrets
+        super.fetch('ns1')
+      end
+    end
+  end
+end
--- a/lib/record_store/provider/ns1/client.rb
+++ b/lib/record_store/provider/ns1/client.rb
@@ -1,0 +1,47 @@
+require 'ns1'
+
+module RecordStore
+  class Provider::NS1 < Provider
+    class Error < StandardError; end
+
+    class Client < ::NS1::Client
+      def initialize(api_key:)
+        super(api_key)
+      end
+
+      def zones
+        super
+      end
+
+      def zone(name)
+        super(name)
+      end
+
+      def record(zone:, fqdn:, type:, must_exist: false)
+        result = super(zone, fqdn, type)
+        raise(Error, result.to_s) if must_exist && result.is_a?(NS1::Response::Error)
+        return nil if result.is_a?(NS1::Response::Error)
+        result
+      end
+
+      def create_record(zone:, fqdn:, type:, params:)
+        result = super(zone, fqdn, type, params)
+        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
+        nil
+      end
+
+      def modify_record(zone:, fqdn:, type:, params:)
+        result = super(zone, fqdn, type, params)
+        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
+        nil
+      end
+
+      def delete_record(zone:, fqdn:, type:)
+        result = super(zone, fqdn, type)
+        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
+        nil
+      end
+    end
+  end
+
+end
--- a/record_store.gemspec
+++ b/record_store.gemspec
@@ -31,6 +31,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'dnsimple', '~> 4.4.0'
   spec.add_runtime_dependency 'google-cloud-dns'
   spec.add_runtime_dependency 'ruby-limiter', '~> 1.0', '>= 1.0.1'
+  spec.add_runtime_dependency 'ns1'
 
 
   spec.add_development_dependency 'byebug'
--- a/record_store.gemspec
+++ b/record_store.gemspec
@@ -32,6 +32,8 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'google-cloud-dns'
   spec.add_runtime_dependency 'ruby-limiter', '~> 1.0', '>= 1.0.1'
 
+
+  spec.add_development_dependency 'byebug'
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'mocha'
--- a/lib/record_store/version.rb
+++ b/lib/record_store/version.rb
@@ -1,3 +1,3 @@
 module RecordStore
-  VERSION = '5.4.3'.freeze
+  VERSION = '5.5.3'.freeze
 end
--- a/template/secrets.json
+++ b/template/secrets.json
@@ -9,6 +9,9 @@
     "account_id": "dnsimple_account_id",
     "api_token": "dnsimple_api_token"
   },
+  "ns1": {
+    "api_key": "ns1_api_key"
+  },
   "google_cloud_dns": {
     "type": "type",
     "project_id": "project_id",
--- a/test/dummy/secrets.json
+++ b/test/dummy/secrets.json
@@ -9,6 +9,9 @@
     "account_id": "1234",
     "api_token": "dnsimple_api_token"
   },
+  "ns1": {
+    "api_key": "ns1_api_key"
+  },
   "google_cloud_dns": {
     "type": "service_account",
     "project_id": "project_id",
--- a/test/fixtures/vcr_cassettes/ns1_add_a_changeset.yml
+++ b/test/fixtures/vcr_cassettes/ns1_add_a_changeset.yml
@@ -1,0 +1,823 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: https://api.nsone.net/v1/zones/test.recordstore.io/test_add_a.test.recordstore.io/A
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      X-Nsone-Key:
+      - "<NS1_API_KEY>"
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 404
+      message: Not Found
+    headers:
+      Date:
+      - Fri, 27 Sep 2019 19:35:13 GMT
+      Content-Type:
+      - application/json
+      Content-Length:
+      - '31'
+      Connection:
+      - keep-alive
+      Set-Cookie:
+      - __cfduid=d1341dbcb81f0839e22d272db47cfbf4c1569612913; expires=Sat, 26-Sep-20
+        19:35:13 GMT; path=/; domain=.nsone.net; HttpOnly
+      X-Ratelimit-Remaining:
+      - '889'
+      X-Ratelimit-By:
+      - customer
+      X-Ratelimit-Limit:
+      - '900'
+      X-Ratelimit-Period:
+      - '300'
+      Pragma:
+      - no-cache
+      Cache-Control:
+      - no-cache
+      Expires:
+      - '0'
+      Expect-Ct:
+      - max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
+      Server:
+      - cloudflare
+      Cf-Ray:
+      - 51cfeb667f66ca98-YYZ
+    body:
+      encoding: UTF-8
+      string: '{"message":"record not found"}
+
+        '
+    http_version: 
+  recorded_at: Fri, 27 Sep 2019 19:35:13 GMT
+- request:
+    method: put
+    uri: https://api.nsone.net/v1/zones/test.recordstore.io/test_add_a.test.recordstore.io/A
+    body:
+      encoding: UTF-8
+      string: '{"answers":[{"answer":["10.10.10.1"]}],"ttl":600,"zone":"test.recordstore.io","domain":"test_add_a.test.recordstore.io","type":"A"}'
+    headers:
+      X-Nsone-Key:
+      - "<NS1_API_KEY>"
+      Content-Type:
+      - application/json
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Date:
+      - Fri, 27 Sep 2019 19:35:14 GMT
+      Content-Type:
+      - application/json
+      Transfer-Encoding:
+      - chunked
+      Connection:
+      - keep-alive
+      Set-Cookie:
+      - __cfduid=d9e435e2448efb348fd9c8b03091816901569612913; expires=Sat, 26-Sep-20
+        19:35:13 GMT; path=/; domain=.nsone.net; HttpOnly
+      X-Ratelimit-Remaining:
+      - '97'
+      X-Ratelimit-By:
+      - customer
+      X-Ratelimit-Limit:
+      - '100'
+      X-Ratelimit-Period:
+      - '200'
+      Pragma:
+      - no-cache
+      Cache-Control:
+      - no-cache
+      - private, max-age=0, no-cache, no-store
+      Expires:
+      - '0'
+      X-Xss-Protection:
+      - 1; mode=block
+      X-Content-Type-Options:
+      - nosniff
+      X-Frame-Options:
+      - SAMEORIGIN
+      Strict-Transport-Security:
+      - max-age=63072000; includeSubdomains; preload
+      Expect-Ct:
+      - max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
+      Server:
+      - cloudflare
+      Cf-Ray:
+      - 51cfeb67d86ccacc-YYZ
+    body:
+      encoding: ASCII-8BIT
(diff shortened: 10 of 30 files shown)


lib/record_store/provider/ns1/client.rb
```diff
@@ -0,0 +1,42 @@
+require 'ns1'
+
+module RecordStore
+  class Provider::NS1 < Provider
+    class Error < StandardError; end
```
XaF: Not a blocker, but could the internal classes be separated with a line to help readability ?

lib/record_store/provider/ns1/client.rb
```diff
@@ -0,0 +1,42 @@
...
+      def record(zone:, fqdn:, type:)
+        result = super(zone, fqdn, type)
+        return nil if result.is_a? NS1::Response::Error
+        result
+      end
+
+      def create_record(zone:, fqdn:, type:, params:)
+        result = super(zone, fqdn, type, params)
+        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
+      end
```
XaF: Maybe return `nil` explicitly at the end so it's clear this won't return anything ?

lib/record_store/provider/ns1/client.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      def modify_record(zone:, fqdn:, type:, params:)
+        result = super(zone, fqdn, type, params)
+        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
+      end
```
XaF: Maybe return `nil` explicitly at the end so it's clear this won't return anything ?

lib/record_store/provider/ns1/client.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      def delete_record(zone:, fqdn:, type:)
+        result = super(zone, fqdn, type)
+        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
+      end
```
XaF: Maybe return `nil` explicitly at the end so it's clear this won't return anything ?

record_store.gemspec
```diff
@@ -31,11 +31,14 @@ Gem::Specification.new do |spec|
...
   spec.add_runtime_dependency 'ruby-limiter', '~> 1.0', '>= 1.0.1'
+  spec.add_runtime_dependency 'ns1'
+
 
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'mocha'
   spec.add_development_dependency 'vcr'
   spec.add_development_dependency 'pry'
+  spec.add_development_dependency 'byebug'
```
XaF: Is this really required ?

.gitignore
```diff
@@ -4,3 +4,4 @@ dev/
 pkg/
 Gemfile.lock
 .bundle/config
+.byebug_history
```
XaF: Is this really something that should be part of the repository ?

lib/record_store.rb
```diff
@@ -31,6 +31,8 @@
 require 'record_store/provider/dynect'
 require 'record_store/provider/dnsimple'
 require 'record_store/provider/google_cloud_dns'
+require 'record_store/provider/ns1/client'
```
XaF: Do we need to require both of them from here, or is the client only required by the provider ? (in which case, we should only require the provider here, and require the client from the provider only)

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
+require 'ns1'
```
XaF: We don't seem to be using the `ns1` gem in here, however we are using the NS1 client built in the other file, this should thus be changed into a `require_relative` of the client

lib/record_store/provider.rb
```diff
@@ -19,6 +19,8 @@ def provider_for(zone_name)
           'DynECT'
         when /googledomains\.com\z/
           'GoogleCloudDNS'
+        when /ns1\.com\z/
```
XaF: Not a blocker at all, and I know this is done the same way above, but wouldn't we want to have the a dot included at the beginning of the matching here? Any domain that would be `anotherns1.com` would match `ns1` too.

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+      def client
+        Provider::NS1::Client.new(api_key: secrets['api_key'])
+      end
+
+      # Downloads all the records from the provider.
+      #
+      # Returns: an array of `Record` for each record in the provider's zone
+      def retrieve_current_records(zone:, stdout: $stdout)
+        records_for_zone(zone)
+         .map{|short_record| build_from_api(short_record, zone)}
```
XaF: Style: the spacing does not seem right here
Could we add a linter to this project to help everyone writing code in it ? https://rubocop.readthedocs.io/en/stable/ could be a solution (it's OSS, and we use it for other internal projects, not sure how complicated it would be to setup for that public project, but pretty sure it's doable)

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+      def records_for_zone(zone)
+        client.zone(zone)["records"]
+      end
+
+      # Creates a new record to the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # record - a kind of `Record`
+      def add(record, zone)
+        new_answers = [{ answer: [ build_api_answer_from_record(record) ]}]
```
XaF: Style: spacing

```ruby
        new_answers = [{ answer: [ build_api_answer_from_record(record) ] }]
```

lib/record_store/provider/ns1/client.rb
```diff
@@ -0,0 +1,42 @@
...
+        super
+      end
+
+      def zone(name)
+        super(name)
+      end
+
+      def record(zone:, fqdn:, type:)
+        result = super(zone, fqdn, type)
+        return nil if result.is_a? NS1::Response::Error
```
XaF: Should we return `nil` without even checking the kind of error ?
Is the absence of record the only kind of error possible to be returned by their API when fetching a record ?

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+          )
+          return
+        end
+
+        existing_answers = existing_record["answers"].map { |answer| symbolize_keys(answer) }
+        client.modify_record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          params: { answers: existing_answers + new_answers }
```
XaF: No update of the TTL ?
How is the TTL handled with many answers ? (taking the shortest ? the longest ? the first we read ?)

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        existing_record = client.record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type
+        )
+        return if existing_record.nil?        
+        
+        pruned_answers = existing_record["answers"]
+          .map { |answer| symbolize_keys(answer) }
+          .reject { |answer| answer[:answer] == build_api_answer_from_record(record) }
```
XaF: This seems a bit weak if we consider that a record might have multiple entries of the same `answer` (with different `meta` for instance). But I guess this is not something to be considered in the `record_store` case as we're only handling static records with very simple setups and no situation where we would need or have `meta` fields. Is that right ?

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+            type: record.type
+          )
+          return
+        end
+
+        client.modify_record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          params: { answers: pruned_answers }
```
XaF: If the TTL is computed, from the answers' TTLs, it would maybe change here ?

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        )
+      end
+
+      # Updates an existing record in the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # id - provider specific ID of record to update
+      # record - a kind of `Record` which the record with `id` should be updated to
+      def update(id, record, zone)
+        super # may not need to be implemented?
```
XaF: Why ?
What is handled by `update` in other providers ?
Why can we ignore it here ?
How can we be sure it will never be called ? If it's the case, we probably would want to raise here instead of calling `super` ?

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+          begin
+            Record.const_get(record_type).new(record)
+          rescue NoMethodError => e
+            raise e
+          end
+        end
+      end
+
+      def build_api_answer_from_record(record)
+        if record.is_a? Record::MX
```
XaF: Style: I usually prefer the form `record.is_a?(Record::MX)` that shows it's a function call

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        elsif record.is_a? Record::CAA
+          [ record.flags, record.tag, record.value ]
+        elsif record.is_a? Record::SRV
+          [ record.priority, record.weight, record.port, record.target ]
+        else
+          [ record.rdata_txt ]
+        end
+      end
+
+      def symbolize_keys(hash)
```
XaF: Don't we have that at a more global project scope ? If not, it might be interesting to have a `utils.rb` file with a `Utils` class that includes that ?

record_store.gemspec
gooallen: Not necessarily, I think it shouldn't be there. Will remove that line!

lib/record_store/provider.rb
cttttt: This is a good point.  Wanna handle this in a follow-on PR, @gooallen?

lib/record_store/provider/ns1.rb
cttttt: In lieu of a linter (which would be something else to handle in a follow-on PR), please provide a suggestion, @XaF .  This is a rare case where Github's border-line-useless single line suggestions would work out 😂

lib/record_store/provider/ns1/client.rb
gooallen: Sure, thanks for catching this! 
It will be updated as below.

```ruby
module RecordStore
  class Provider::NS1 < Provider
    class Error < StandardError; end

    class Client < ::NS1::Client
    (snip)
```

lib/record_store/provider/ns1/client.rb
gooallen: Sure, it will be much clear that it won't return anything with explicit return `nil`.

```ruby
      def create_record(zone:, fqdn:, type:, params:)
        result = super(zone, fqdn, type, params)
        raise(Error, result.to_s) if result.is_a? NS1::Response::Error
        return nil
      end
```

lib/record_store/provider/ns1/client.rb
gooallen: Same as above

lib/record_store/provider/ns1/client.rb
gooallen: Same as above

lib/record_store/provider.rb
gooallen: @cttttt Will look into it.

record_store.gemspec
cttttt: Consider keeping it to pay it forward to future maintainers who need to pop into a debugger.  This is probably why `pry` was already present.  To @XaF 's point, though, we should never check in code that *`require`s* `byebug`.

.gitignore
XaF: My bad, I realize this is the `.gitignore` so we don't really care. Will resolve this.

record_store.gemspec
XaF: I did not realize it was part of `add_development_dependency` as the diff did not show the lines above in the code.
You can keep this around and ignore my comment, I'll resolve.

lib/record_store/provider/ns1.rb
XaF:
```suggestion
         .map { |short_record| build_from_api(short_record, zone) }
```

lib/record_store/provider/ns1.rb
XaF: Updated my code as a suggestion (had to google to find out how.... and this is as easy as saying the code is of `suggestion` type...)
Followed what rubocop does for `spy`: no spaces in list, spaces in hash

lib/record_store/provider/ns1.rb
XaF:
```suggestion
        if record.is_a?(Record::MX)
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+          rescue NoMethodError => e
+            raise e
+          end
+        end
+      end
+
+      def build_api_answer_from_record(record)
+        if record.is_a? Record::MX
+          [ record.preference, record.exchange ]
+        elsif record.is_a? Record::TXT or record.is_a? Record::SPF
```
XaF:
```suggestion
        elsif record.is_a?(Record::TXT) or record.is_a?(Record::SPF)
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+          [ record.txtdata ]
+        elsif record.is_a? Record::CAA
```
XaF:
```suggestion
        elsif record.is_a?(Record::CAA)
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+          [ record.flags, record.tag, record.value ]
+        elsif record.is_a? Record::SRV
```
XaF:
```suggestion
        elsif record.is_a?(Record::SRV)
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+            Record.const_get(record_type).new(record)
+          rescue NoMethodError => e
+            raise e
+          end
+        end
+      end
+
+      def build_api_answer_from_record(record)
+        if record.is_a? Record::MX
+          [ record.preference, record.exchange ]
```
XaF:
```suggestion
          [record.preference, record.exchange]
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        elsif record.is_a? Record::TXT or record.is_a? Record::SPF
+          [ record.txtdata ]
```
XaF:
```suggestion
          [record.txtdata]
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        elsif record.is_a? Record::CAA
+          [ record.flags, record.tag, record.value ]
```
XaF:
```suggestion
          [record.flags, record.tag, record.value]
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        elsif record.is_a? Record::SRV
+          [ record.priority, record.weight, record.port, record.target ]
```
XaF:
```suggestion
          [record.priority, record.weight, record.port, record.target]
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+        else
+          [ record.rdata_txt ]
```
XaF:
```suggestion
          [record.rdata_txt]
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+      def build_api_answer_from_record(record)
+        if record.is_a? Record::MX
+          [ record.preference, record.exchange ]
+        elsif record.is_a? Record::TXT or record.is_a? Record::SPF
+          [ record.txtdata ]
+        elsif record.is_a? Record::CAA
+          [ record.flags, record.tag, record.value ]
+        elsif record.is_a? Record::SRV
+          [ record.priority, record.weight, record.port, record.target ]
+        else
```
XaF: I don't know how it's done in other providers, but I would probably go for a `elsif record.is_a?(Record::TXT)` for that, and use the `else` to raise an exception for an unhandled case reaching that provider

lib/record_store/provider/ns1.rb
gooallen: Great,
it will be updated like this!
```ruby
require_relative 'ns1'

module RecordStore
  class Provider::NS1 < Provider
  (snip)
```

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested, thanks :)

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested, thanks :)

lib/record_store/provider/ns1.rb
gooallen: That's right. But I can definitely test with some complicated setups that you can think of.

lib/record_store/provider/ns1.rb
XaF: You mean `require_relative 'ns1/client'` ? Or does `require_relative 'ns1'` include automatically everything that's in the `ns1` directory ? (Ruby noob :) )

lib/record_store/provider/ns1.rb
gooallen: We found out that some of providers are not even using `update` method.
But it turned out that it does need to be implemented `update` method. :(
It will be updated in a new PR soon.

lib/record_store/provider/ns1.rb
gooallen: It's been updated as you suggested!

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,192 @@
...
+            priority, weight, port, host = answer.split(' ')
+
+            record.merge!(
+              priority: priority.to_i,
+              weight: weight.to_i,
+              port: port.to_i,
+              target: Record.ensure_ends_with_dot(host),
+            )
+          end
+          begin
```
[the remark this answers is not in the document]
gooallen: It seems that we don't really need to put `try / catch` block here. (other providers also have no `try / catch`. This has been updated as below.

```ruby
          end
          Record.const_get(record_type).new(record)
        end
      end
```

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
XaF: My idea would be that if there is a chance we need to handle the meta thing, then yes, let's take time for this. But if it's not something that can happen with the record store in the current state, we can ignore this for now (as anyway, in this case, supporting that case would require more work elsewhere, and the addition to this provider could be done at the same time).

So my question to you would be: is this a case that can happen (having multiple identical answers with different metas) with the current state of the record store ?

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: It has been updated as you suggested!

lib/record_store/provider/ns1.rb
gooallen: Other providers use all `elsif`s and no raise an exception but your comment should be better option to deal with an unhandled case. Let me try it. Thanks.

lib/record_store.rb
gooallen: I might be wrong but I believe we need both of them from here.

XaF: If you require `record_store/provider/ns1` and `record_store/provider/ns1.rb` requires `record_store/provider/ns1/client`, then the requirement is already there, just indirect.
It "hides" the fact that that "client" level exists, as the `record_store.rb` file does not care about it. This works mostly because `record_store.rb` uses nothing that's in `ns1/client.rb` but only things in `ns1.rb`. The one using `ns1/client.rb` is `ns1.rb`.

If A requires B and B requires C, indirectly, A will require C, so we're still good, but it's cleaner :)

lib/record_store/provider/ns1.rb
cttttt: I think for now, we should assume that record store will the exclusive maintainer of the zones it's touching.  If we don't allow it to create complicated annotated records, it shouldn't be expected to maintain them.

If we unleash record store on a zone that has already been populated, ... the behaviour should be undefined.  To tighten this up, I'd go so far as blacklisting fields on answers.  If an answer has a `meta` field, `raise` because record_store must not have created it.

lib/record_store/provider/ns1.rb
cttttt: This is an interesting point.  It's impossible to have, say, two A records for a single fqdn with different TTLs.  We'll want to be deliberate about how to handle attempts to change a TTL.  We could either raise when inconsistent TTLs are encountered (probably a smart choice) or update the TTL to that of the latest record for an FQDN and type.

lib/record_store/provider/ns1.rb
gooallen: I see what you mean! Will be updated as suggested (Ruby noob here :) )

lib/record_store/provider/ns1/client.rb
gooallen: It seems weird but if you check `error.rb` of NS1 library, you might understand why we ended up approaching this way. 
https://github.com/kitop/ns1-ruby/blob/master/lib/ns1/response/error.rb

lib/record_store/provider/ns1.rb
gooallen: Ah, I left a comment here but I guess it's all gone without saving once I canceled review mode or something (all comments were pending). I left the comment with similar idea of Chris here. I also quickly tested with inconsistent TTLs, and the TTL used for the first record always stays.

lib/record_store/provider/ns1.rb
gooallen: I think this questions related to what we are discussing about TTL now. Many answers will be adding to the record with the TTL which was set for the creation of the record. So I could say, the TTL will be handled with the first we read from given questions.

lib/record_store.rb
gooallen: Great! Thanks for the details! Yeah, it all makes sense, much cleaner now.

lib/record_store/provider/ns1.rb
gooallen: I am not 100% getting what the `global project scope` means here. Could you give me more details?

lib/record_store/provider/ns1.rb
cttttt: While working through this, we came to the following conclusions:

- If the provider's asked to `add` or `update` a record, it will have been provided a desired `RecordStore::Record` (complete with desired TTL), and will therefore be able to set the TTL of the record in NS1 to at least some desired TTL for records with that fqdn and type.
- If the provider's asked to remove a record, it will only have been provided the `RecordStore::Record` (complete with TTL) for a record that we want to discard.  There isn't enough information to know any of the other desired TTLs.  So, if someone removes one record from an answer set, the TTL will remain unchanged.  We'll have to figure out if this a huge problem or not going forward, but...it's kinda a corner case.

lib/record_store/provider/ns1.rb
cttttt: We settled on not trying to go all fully-typed here, only doing type checking for exceptional cases.

lib/record_store/provider/ns1.rb
cttttt: There's a symbolize_keys in [ActiveSupport](http://api.rubyonrails.org/classes/Hash.html#method-i-symbolize_keys), however it doesn't feel right pulling in the entire library for one method.  We didn't see any similar methods in RecordStore, so decided to implement it, and keep it in the smallest scope that encompasses all of its uses.  If it ends up needing to be used elsewhere, it can moved to a larger scope.

cttttt:
@XaF , please have another look.

lib/record_store/provider/ns1.rb
cttttt: Turned out that providers were using the `update` method 😢 , so we needed to run in and implement it.  Fun stuff.

lib/record_store/provider/ns1/client.rb
cttttt: We ended up enhancing record with an option that allows it to return nil for non-existent records.

lib/record_store/provider.rb
```diff
@@ -19,6 +19,8 @@ def provider_for(zone_name)
           'DynECT'
         when /googledomains\.com\z/
           'GoogleCloudDNS'
+        when /.nsone\.net\z/
```
XaF: if using the `.`, please backslash-it !
```suggestion
        when /\.nsone\.net\z/
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+            must_exist: true,
+          )
+        end
+
+        full_api_records.map { |r| build_from_api(r, zone) }.flatten
+      end
+
+      # Returns an array of the zones managed by provider as strings
+      def zones
+        client.zones.map do |zone|
```
XaF:
```suggestion
        client.zones.map { |zone| zone['zone'] }
```

Would probably be sufficient here (style preference here :D )

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+      end
+
+      # Creates a new record to the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # record - a kind of `Record`
+      def add(record, zone)
+        new_answers = [{ answer: build_api_answer_from_record(record) }]
+
+        record_fqdn = record.fqdn.sub(/.$/, '')
```
XaF: This will remove any character at the end of the string ? Was it supposed to match the dot (`.`) ? In which case, `\.` would be better, or even nicer to avoid regex for nothing:
```suggestion
        record_fqdn = record.fqdn.delete_suffix('.')
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+          client.create_record(
+            zone: zone,
+            fqdn: record_fqdn,
+            type: record.type,
+            params: { answers: new_answers, ttl: record.ttl }
+          )
+          return
+        end
+
+        existing_answers = existing_record["answers"].map { |answer| symbolize_keys(answer) }
```
XaF:
```suggestion
        existing_answers = existing_record['answers'].map(&:symbolize_keys)
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+          params: { answers: existing_answers + new_answers, ttl: record.ttl }
+        )
+      end
+
+      # Deletes an existing record from the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # record - a kind of `Record`
+      def remove(record, zone)
+        record_fqdn = record.fqdn.sub(/.$/, '')
```
XaF:
```suggestion
        record_fqdn = record.fqdn.delete_suffix('.')
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+
+        existing_record = client.record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type
+        )
+        return if existing_record.nil?
+
+        pruned_answers = existing_record["answers"]
+          .map { |answer| symbolize_keys(answer) }
```
XaF:
```suggestion
          .map(&:symbolize_keys)
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+        record_fqdn = record.fqdn.sub(/.$/, '')
+
+        existing_record = client.record(
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type
+        )
+        return if existing_record.nil?
+
+        pruned_answers = existing_record["answers"]
```
XaF: Personal preference: avoid double quotes unless when interpolating

```suggestion
        pruned_answers = existing_record['answers']
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+        )
+      end
+
+      # Updates an existing record in the zone. It is expected this call modifies external state.
+      #
+      # Arguments:
+      # id - provider specific ID of record to update
+      # record - a kind of `Record` which the record with `id` should be updated to
+      def update(id, record, zone)
+        record_fqdn = record.fqdn.sub(/.$/, '')
```
XaF:
```suggestion
        record_fqdn = record.fqdn.delete_suffix('.')
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+          zone: zone,
+          fqdn: record_fqdn,
+          type: record.type,
+          must_exist: true,
+        )
+
+        # Identify the answer in this record with the matching ID, and update it
+        updated = false
+        existing_record["answers"].each do |answer|
+          if answer["id"] == id
```
XaF: To gain an indentation level, you can revert the condition:

```suggestion
          next if answer["id"] != id
```

I would also add a `break` at the end of the loop to avoid checking all of the other records, as we only want the one matching the id.

Another possibility (probably even more ruby-like) would be:

```ruby
existing_record['answers'].select { |answer| answer['id'] == id }.each do |answer|
  updated = true
  answer['answer'] = build_api_answer_from_record(record)
end
```

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+
+        record_type = api_record["type"]
+        api_record["answers"].map do |api_answer|
+          answer = api_answer["answer"]
+          record = {
+            ttl: api_record["ttl"],
+            fqdn: fqdn.downcase,
+            record_id: api_answer["id"]
+          }
+          return if record_type == 'SOA'
```
XaF: If you want to exit because of the record type, why doing it here in the `map` instead of just after reading the `record_type` value before the loop ?

lib/record_store/provider/ns1.rb
```diff
@@ -0,0 +1,222 @@
...
+
+            record.merge!(
+              flags: flags.to_i,
+              tag: tag,
+              value: Record.unquote(value),
+            )
+          when 'CNAME'
+            record.merge!(cname: answer.first)
+          when 'MX'
+
```
XaF: Extra line

lib/record_store/provider/ns1.rb
cttttt: `symbolize_keys` is unfortunately not a method of the `map`-ed type, so this won't work out.

lib/record_store/provider/ns1.rb
cttttt:
```suggestion
        existing_answers = existing_record['answers'].map { |answer| symbolize_keys(answer) }
```

PR #91, opened 2019-10-02
Style: trailing comma in hash literal

gooallen:
**Style: trailing comma in hash literal**

Put a comma after the last item of a multiline hash.

### DIFF
--- a/test/providers/dynect_test.rb
+++ b/test/providers/dynect_test.rb
@@ -16,7 +16,7 @@ class DynECTTest < Minitest::Test
       'record_type' => 'A',
       'ttl' => 600,
       'rdata' => {
-        'address' => '10.11.12.13'
+        'address' => '10.11.12.13',
       }
     )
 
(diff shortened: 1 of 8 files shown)


test/record_test.rb
```diff
@@ -27,7 +27,7 @@ class RecordTest < Minitest::Test
       port: 80,
       target: 'target-srv.dns-test.shopify.io.',
       ttl: 60
```
XaF: Interesting it's not complaining about this

test/providers/dynect_test.rb
```diff
@@ -16,7 +16,7 @@ def test_build_a_from_api
       'record_type' => 'A',
       'ttl' => 600,
       'rdata' => {
-        'address' => '10.11.12.13'
+        'address' => '10.11.12.13',
       }
```
XaF: Interesting it's not complaining about this

PR #94, opened 2019-10-02
Style: zero length predicate

gooallen:
**Style: zero length predicate**

Use !empty? instead of size > 0.

### DIFF
--- a/lib/record_store/cli.rb
+++ b/lib/record_store/cli.rb
@@ -210,7 +210,7 @@ module RecordStore
         end
       end
 
-      if !invalid_zones.size.empty?
+      unless invalid_zones.empty?
         abort "The following zones were invalid: #{invalid_zones.join(', ')}"
       else
         puts "All zones have valid definitions."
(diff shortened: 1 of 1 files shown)


lib/record_store/cli.rb
```diff
@@ -200,7 +200,7 @@ def validate_records
         end
 
         invalid_records = zone.records.reject(&:valid?)
-        puts '  Invalid records' if invalid_records.size > 0
+        puts '  Invalid records' if !invalid_records.size.empty?
```
XaF:
```suggestion
        puts '  Invalid records' unless invalid_records.size.empty?
```

lib/record_store/cli.rb
```diff
@@ -210,7 +210,7 @@ def validate_records
         end
       end
 
-      if invalid_zones.size > 0
+      if !invalid_zones.size.empty?
```
XaF:
```suggestion
      unless invalid_zones.size.empty?
```

Or even better, so it's nicer to read, we can revert the condition:

```ruby
      if invalid_zones.size.empty?
        puts 'All zones have valid definitions.'
      else
        abort "The following zones were invalid: #{invalid_zones.join(', ')}"
      end
```

PR #97, opened 2019-10-03
Style: update using ! instead of not

gooallen:
Style: update using ! instead of not

### DIFF
--- a/lib/record_store/zone.rb
+++ b/lib/record_store/zone.rb
@@ -47,7 +47,7 @@ module RecordStore
         (1..MAX_PARALLEL_THREADS).map do
           Thread.new do
             current_zone = nil
-            while !zones.empty?
+            while zones.any? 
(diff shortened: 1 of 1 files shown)


lib/record_store/zone.rb
```diff
@@ -47,7 +47,7 @@ def modified(verbose: false)
         (1..MAX_PARALLEL_THREADS).map do
           Thread.new do
             current_zone = nil
-            while not zones.empty?
+            while !zones.empty?
```
XaF: I think you can do with
```suggestion
            until zones.empty?
```

Or with

```suggestion
            while zones.any? 
```

Please verify that I'm not lying though (you can run `irb` in your console or just create a basic ruby file to test)

gooallen: Of course, I believe you are not lying :)
I also confirmed that your suggestion works in the same way.
Thanks!

Issue #84, opened 2019-10-01
Add  rubocop support for record_store


gooallen:
Allow to have code linting for the `record_store` by enabling rubocop.

This will require:
1/ Fix all the issues that might trigger rubocop in the project
2/ Submit a PR with the rubocop configuration for any future change to be merged in the project

PR #104, opened 2019-10-03
Style: next

gooallen:
Related to #84
 - use next to skip iteration

### DIFF
--- a/lib/record_store/cli.rb
+++ b/lib/record_store/cli.rb
@@ -83,7 +83,7 @@ module RecordStore
             end
           end
 
-          if options.fetch('verbose')
+          next if options.fetch('verbose')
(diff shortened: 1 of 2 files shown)


lib/record_store/cli.rb
```diff
@@ -83,7 +83,7 @@ def diff
             end
           end
 
-          if options.fetch('verbose')
+          next if options.fetch('verbose')
```
XaF: Pretty sure what you want is actually:

```
next unless options.fetch('verbose')

puts "Unchanged:"
changeset.unchanged.each do |record|	            changeset.unchanged.each do |record|
  puts " - #{record}"	              puts " - #{record}"
end	            end
```

PR #109, opened 2019-10-03
Layout: multiline method call brace layout

gooallen:
Related to #84

### DIFF
--- a/test/providers/dynect_test.rb
+++ b/test/providers/dynect_test.rb
@@ -10,14 +10,16 @@ class DynECTTest < Minitest::Test
   end
 
   def test_build_a_from_api
-    record = Provider::DynECT.send(:build_from_api,
+    record = Provider::DynECT.send(
+      :build_from_api,
       'zone' => 'dns-test.shopify.io',
       'fqdn' => 'test.dns-test.shopify.io',
       'record_type' => 'A',
       'ttl' => 600,
       'rdata' => {
         'address' => '10.11.12.13',
-      })
+      }
+    )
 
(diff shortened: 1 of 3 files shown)


test/providers/dynect_test.rb
```diff
@@ -204,8 +195,7 @@ def test_build_txt_from_api_handles_mixed_case_fqdn
       "record_type" => "TXT",
       "rdata" => {
         "txtdata" => "Hello, world!",
-      },
-    )
+      },)
```
XaF: This feels weird to have the comma in 'sandwich' between closing chars.
If you fix the first line instead (as suggested above) this will be fine

test/providers/dynect_test.rb
```diff
@@ -17,8 +17,7 @@ def test_build_a_from_api
       'ttl' => 600,
       'rdata' => {
         'address' => '10.11.12.13',
-      }
-    )
+      })
```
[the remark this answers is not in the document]
gooallen: This is good! Thanks!

test/providers/google_cloud_dns_test.rb
```diff
@@ -16,8 +16,7 @@ def test_build_a_from_api
         'A',
         600,
         '10.11.12.13'
-      )
-    )
+      ))
```
XaF: Please do the same in that file! (first parameter of the function in another line instead of merging the closing chars)

gooallen: This file has been updated :)

PR #136, opened 2019-10-05
Lint: debugger

gooallen:
Related to #84
 - remove debugger entry point


bin/console
```diff
@@ -8,4 +8,3 @@ RecordStore.zones_path = File.expand_path('../../dev/zones', __FILE__)
 RecordStore.config_path = File.expand_path('../../dev/config.yml', __FILE__)
 
 require 'pry'
-binding.pry(RecordStore)
```
XaF: Mh given the file it's in, I think I would go with the "ignore the error for this" instead
Same as we did for the interface, add a comment to ask rubocop to ignore that error in that line!

PR #137, opened 2019-10-05
Lint: unused method argument

gooallen:
Related to #84

### DIFF
--- a/lib/record_store/provider/google_cloud_dns.rb
+++ b/lib/record_store/provider/google_cloud_dns.rb
@@ -3,7 +3,7 @@ require 'google/cloud/dns'
 module RecordStore
   class Provider::GoogleCloudDNS < Provider
     class << self
-      def apply_changeset(changeset, stdout = $stdout)
+      def apply_changeset(changeset, stdout = $stdout) # rubocop:disable Lint/UnusedMethodArgument
         zone = session.zone(convert_to_name(changeset.zone))
 
         deletions = convert_records_to_gcloud_record_sets(zone, changeset.current_records)
--- a/lib/record_store/provider/google_cloud_dns.rb
+++ b/lib/record_store/provider/google_cloud_dns.rb
@@ -3,7 +3,7 @@ require 'google/cloud/dns'
 module RecordStore
   class Provider::GoogleCloudDNS < Provider
     class << self
-      def apply_changeset(changeset, stdout = $stdout) # rubocop:disable Lint/UnusedMethodArgument
+      def apply_changeset(changeset, _stdout)
         zone = session.zone(convert_to_name(changeset.zone))
 
         deletions = convert_records_to_gcloud_record_sets(zone, changeset.current_records)
--- a/lib/record_store/provider/google_cloud_dns.rb
+++ b/lib/record_store/provider/google_cloud_dns.rb
@@ -3,7 +3,7 @@ require 'google/cloud/dns'
 module RecordStore
   class Provider::GoogleCloudDNS < Provider
     class << self
-      def apply_changeset(changeset, _stdout)
+      def apply_changeset(changeset, _stdout = nil)
         zone = session.zone(convert_to_name(changeset.zone))
 
         deletions = convert_records_to_gcloud_record_sets(zone, changeset.current_records)
--- a/lib/record_store/provider/ns1.rb
+++ b/lib/record_store/provider/ns1.rb
@@ -12,7 +12,7 @@ module RecordStore
       # Downloads all the records from the provider.
       #
       # Returns: an array of `Record` for each record in the provider's zone
-      def retrieve_current_records(zone:, stdout: $stdout)
+      def retrieve_current_records(zone:, stdout: $stdout) # rubocop:disable Lint/UnusedMethodArgument
         full_api_records = records_for_zone(zone).map do |short_record|
           client.record(
             zone: zone,
@@ -129,7 +129,11 @@ module RecordStore
           answer["answer"] = build_api_answer_from_record(record)
         end
 
-        raise(Error, "while trying to update a record, could not find answer with fqdn: #{record.fqdn}, type; #{record.type}, id: #{id}") unless updated
+        unless updated
+          error = +'while trying to update a record, could not find answer with fqdn: '
+          error << "#{record.fqdn}, type; #{record.type}, id: #{id}"
+          raise Error, error
+        end
 
         client.modify_record(
           zone: zone,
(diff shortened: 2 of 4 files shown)


lib/record_store/provider/google_cloud_dns.rb
```diff
@@ -3,7 +3,7 @@
 module RecordStore
   class Provider::GoogleCloudDNS < Provider
     class << self
-      def apply_changeset(changeset, stdout = $stdout)
+      def apply_changeset(changeset, stdout = $stdout) # rubocop:disable Lint/UnusedMethodArgument
```
XaF: This is not an interface here, please rename the parameter `stdout` in `_stdout` (and do not set default value as we don't use it anyway) instead of ignoring the error!

I'm also confused by the interface of that class, where we sometimes have ordered parameters and sometimes named ones... But in this case we could just rename and remove default as not used (which is not possible when using named parameters as it would change the behavior)

lib/record_store/provider/ns1.rb
```diff
@@ -139,7 +139,7 @@ def update(id, record, zone)
         )
       end
 
-      def build_from_api(api_record, zone)
+      def build_from_api(api_record, zone) # rubocop:disable Lint/UnusedMethodArgument
```
XaF: This is not an interface here, please rename the parameter `zone` in `_zone` instead of ignoring the error!

Also, side note but, I'm surprised we don't have an error checking verifying that the `api_record['domain']` corresponds to the `zone` we received. That shouldn't happen, but if it does we just act as if nothing here ? Maybe @cttttt would have more context on this ?

lib/record_store/zone.rb
```diff
@@ -41,7 +41,7 @@ def filter_records(current_records, ignore_patterns)
       end
 
       MAX_PARALLEL_THREADS = 10
-      def modified(verbose: false)
+      def modified(verbose: false) # rubocop:disable Lint/UnusedMethodArgument
```
XaF: Keep it as-is here, as renaming the parameter might have an impact...

lib/record_store/provider/ns1.rb
cttttt: - `zone` isn't used here and this should actually a private method.  The parameter can probably be removed.  I think this was modelled off of similar private methods in other providers, which used `zone` because record domains in the API weren't fqdns.  The NS1 API returns fqdns as record domains.

lib/record_store/provider/google_cloud_dns.rb
```diff
@@ -3,7 +3,7 @@
 module RecordStore
   class Provider::GoogleCloudDNS < Provider
     class << self
-      def apply_changeset(changeset, stdout = $stdout)
+      def apply_changeset(changeset, _stdout)
```
gooallen: Complaining here from `rake test` @XaF 
```
ArgumentError: wrong number of arguments (given 1, expected 2)
    /Users/byeongyun/src/github.com/Shopify/record_store/lib/record_store/provider/google_cloud_dns.rb:6:in `apply_changeset'
```

XaF: Huh, my bad. Add a ` = nil` for the arg (as anyway you're not using it)

gooallen: Hmm, got another error after `a = nil`
Seems like it's used in NS1.rb 

15 errors have similar looking.
```
 15) Error:
NS1Test#test_txt_record_retrieved_after_adding_record_changeset:
ArgumentError: wrong number of arguments (given 2, expected 1)
    /Users/byeongyun/src/github.com/Shopify/record_store/lib/record_store/provider/ns1.rb:142:in `build_from_api'
    /Users/byeongyun/src/github.com/Shopify/record_store/lib/record_store/provider/ns1.rb:25:in `block in retrieve_current_records'
    /Users/byeongyun/src/github.com/Shopify/record_store/lib/record_store/provider/ns1.rb:25:in `map'
    /Users/byeongyun/src/github.com/Shopify/record_store/lib/record_store/provider/ns1.rb:25:in `retrieve_current_records'
    /Users/byeongyun/src/github.com/Shopify/record_store/test/providers/ns1_test.rb:430:in `block in test_txt_record_retrieved_after_adding_record_changeset'
    /Users/byeongyun/.gem/ruby/2.4.3/gems/vcr-5.0.0/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
    /Users/byeongyun/.gem/ruby/2.4.3/gems/vcr-5.0.0/lib/vcr.rb:188:in `use_cassette'
    /Users/byeongyun/src/github.com/Shopify/record_store/test/providers/ns1_test.rb:423:in `test_txt_record_retrieved_after_adding_record_changeset'
```

XaF: Wait, did you add `a = nil` ? I meant `= nil` after `_stdout`:

```suggestion
      def apply_changeset(changeset, _stdout = nil)
```

i.e. no need to set an useful value like `$stdout` as we're not doing anything with it anyway

gooallen: Hmm, I added it after `_stdout` like you described above. @XaF

XaF: Are you sure this is not related to another change somewhere else ? Because that traceback above talks about `build_from_api` and not `apply_changeset`. Probably related to the removal of the `zone` argument for `build_from_api`, that was actually passed by the `retrieve_current_records` method: `full_api_records.map { |r| build_from_api(r, zone) }.flatten.compact`

Please replace `full_api_records.map { |r| build_from_api(r, zone) }.flatten.compact` by `full_api_records.map { |r| build_from_api(r) }.flatten.compact` to solve that issue.

gooallen: Thanks for your help. That works fine!

lib/record_store/provider/ns1.rb
```diff
@@ -12,7 +12,7 @@ def client
       # Downloads all the records from the provider.
       #
       # Returns: an array of `Record` for each record in the provider's zone
-      def retrieve_current_records(zone:, stdout: $stdout) # rubocop:disable Lint/UnusedMethodArgument
+      def retrieve_current_records(zone:, stdout: $stdout)
```
XaF: Wait I'm confused, it's still not using stdout?

XaF: Approving but please check this is not raising an error :)

XaF: You are using `zone` so why rename it here?
But you are not using `stdout`, and as explained previously, as it's a named parameter, you cannot really rename it if it's a method called for many providers. The only possibility here is to add the linter ignore as you had previously

gooallen: Thanks for the quick comment!

PR #140, opened 2019-10-07
Metrics: line length

gooallen:
Related to #84
 - line is too long (over 120)

### DIFF
--- a/lib/record_store.rb
+++ b/lib/record_store.rb
@@ -46,7 +46,8 @@ module RecordStore
     end
 
     def zones_path
-      @zones_path ||= Pathname.new(File.expand_path(config.fetch('zones_path'), File.dirname(config_path))).realpath.to_s
+      @zones_path ||= Pathname.new(File.expand_path(config.fetch('zones_path'), \
+        File.dirname(config_path))).realpath.to_s
     end
 
     def config_path
--- a/lib/record_store.rb
+++ b/lib/record_store.rb
@@ -49,7 +49,7 @@ module RecordStore
       @zones_path ||= Pathname.new(
         File.expand_path(config.fetch('zones_path'),
         File.dirname(config_path)),
-        ).realpath.to_s
+      ).realpath.to_s
     end
 
     def config_path
--- a/lib/record_store.rb
+++ b/lib/record_store.rb
@@ -46,8 +46,10 @@ module RecordStore
     end
 
     def zones_path
-      @zones_path ||= Pathname.new(File.expand_path(config.fetch('zones_path'), \
-        File.dirname(config_path))).realpath.to_s
+      @zones_path ||= Pathname.new(
+        File.expand_path(config.fetch('zones_path'),
+        File.dirname(config_path)),
+        ).realpath.to_s
     end
 
     def config_path
--- a/lib/record_store/cli.rb
+++ b/lib/record_store/cli.rb
@@ -226,8 +226,10 @@ module RecordStore
         end
 
         unless removals.empty?
-          abort "As a safety measure, you cannot remove more than #{MAXIMUM_REMOVALS} "\
-                "records at a time per zone. (zones failing this: #{removals.map(&:name).join(', ')})"
+          error = +"As a safety measure, you cannot remove more than #{MAXIMUM_REMOVALS} "
+          error << 'records at a time per zone. '
+          error << "(zones failing this: #{removals.map(&:name).join(', ')})"
+          abort error
         end
       end
     end
--- a/lib/record_store/cli.rb
+++ b/lib/record_store/cli.rb
@@ -113,7 +113,8 @@ module RecordStore
 
     option :name, desc: 'Zone to download', aliases: '-n', type: :string, required: true
     option :provider, desc: 'Provider in which this zone exists', aliases: '-p', type: :string
-    desc 'download', 'Downloads all records from zone and creates YAML zone definition in zones/ e.g. record-store download --name=shopify.io'
+    desc 'download', 'Downloads all records from zone and creates YAML zone definition in zones/ '\
+                     'e.g. record-store download --name=shopify.io'
     def download
       name = options.fetch('name')
       abort 'Please omit the period at the end of the zone' if name.ends_with?('.')
@@ -151,7 +152,8 @@ module RecordStore
       abort "Please omit the period at the end of the zone" if name.ends_with?('.')
 
       yaml = YAML.load_file("#{RecordStore.zones_path}/#{name}.yml")
-      yaml.fetch(name).fetch('records').sort_by! { |r| [r.fetch('fqdn'), r.fetch('type'), r['nsdname'] || r['address']] }
+      yaml.fetch(name).fetch('records').sort_by! \
+      { |r| [r.fetch('fqdn'), r.fetch('type'), r['nsdname'] || r['address']] }
 
       File.write("#{RecordStore.zones_path}/#{name}.yml", yaml.deep_stringify_keys.to_yaml.gsub("---\n", ''))
     end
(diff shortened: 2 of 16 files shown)


lib/record_store/cli.rb
```diff
@@ -224,7 +226,8 @@ def validate_change_size
         end
 
         unless removals.empty?
-          abort "As a safety measure, you cannot remove more than #{MAXIMUM_REMOVALS} records at a time per zone. (zones failing this: #{removals.map(&:name).join(', ')})"
+          abort "As a safety measure, you cannot remove more than #{MAXIMUM_REMOVALS} "\
```
XaF: Is this supposed to be aligned exactly the same way ?
Maybe you could use a temporary variable to store the string for readability:

```
error = +"As a safety measure, you cannot remove more than #{MAXIMUM_REMOVALS} "
error << 'records at a time per zone. '
error << "(zones failing this: #{removals.map(&:name).join(', ')})"
abort error
```

lib/record_store/provider/dynect.rb
```diff
@@ -34,7 +34,8 @@ def apply_changeset(changeset, stdout = $stdout)
 
       # returns an array of Record objects that match the records which exist in the provider
       def retrieve_current_records(zone:, stdout: $stdout)
-        session.get_all_records(zone).body.fetch('data').flat_map do |type, records|
+        session.get_all_records(zone).body.fetch('data').flat_map\
```
XaF: I don't think this will be needed in the refactor if you rename the variable with a `_` prefix instead of ignoring the rubocop error here.

lib/record_store/provider/ns1.rb
```diff
@@ -129,7 +129,8 @@ def update(id, record, zone)
           answer["answer"] = build_api_answer_from_record(record)
         end
 
-        raise(Error, "while trying to update a record, could not find answer with fqdn: #{record.fqdn}, type; #{record.type}, id: #{id}") unless updated
+        raise(Error, "while trying to update a record, could not find answer with fqdn: "\
```
XaF: I don't really like that split, maybe...

```
unless updated
  error = +'while trying to update a record, could not find answer with fqdn: '
  error << "#{record.fqdn}, type; #{record.type}, id: #{id}"
  raise Error, error
end
```

lib/record_store/record/alias.rb
```diff
@@ -2,7 +2,11 @@ module RecordStore
   class Record::ALIAS < Record
     attr_accessor :alias
 
-    validates :alias, presence: true, format: { with: Record::CNAME_REGEX, message: 'is not a fully qualified domain name' }
+    validates :alias, presence: true, format:
```
XaF: Maybe, just maybe, I would put the opening of the bracket on the same line as `format`, as it would be clearer that it's a follow-up to that line. Would rubocop complain about that ?

lib/record_store/record/caa.rb
```diff
@@ -5,7 +5,11 @@ class Record::CAA < Record
     LABEL_REGEX = '[a-z0-9](?:-*[a-z0-9])*'
     DOMAIN_REGEX = /\A#{LABEL_REGEX}(?:\.#{LABEL_REGEX})\z/i
 
-    validates :flags, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than_or_equal_to: 255 }, presence: true
+    validates :flags, numericality:
```
XaF: Would suggest to move `, presence: true` before the `numericality:` key, and as above, keep the `{` on the same line as `numericality:`

lib/record_store/record/cname.rb
```diff
@@ -2,7 +2,11 @@ module RecordStore
   class Record::CNAME < Record
     attr_accessor :cname
 
-    validates :cname, presence: true, format: { with: Record::CNAME_REGEX, message: 'is not a fully qualified domain name' }
+    validates :cname, presence: true, format:
```
XaF: Same for `{` if that works

lib/record_store/record/mx.rb
```diff
@@ -2,8 +2,16 @@ module RecordStore
   class Record::MX < Record
     attr_accessor :exchange, :preference
 
-    validates :preference, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, presence: true
-    validates :exchange, presence: true, format: { with: Record::FQDN_REGEX, message: 'is not a fully qualified domain name' }
+    validates :preference, numericality:
```
XaF: Same for `presence: true` and for `{`

lib/record_store/record/mx.rb
```diff
@@ -2,8 +2,16 @@ module RecordStore
...
+      {
+        only_integer: true,
+        greater_than_or_equal_to: 0,
+      }, presence: true
+    validates :exchange, presence: true, format:
```
XaF: Same for `{`

lib/record_store/record/ns.rb
```diff
@@ -2,7 +2,11 @@ module RecordStore
   class Record::NS < Record
     attr_accessor :nsdname
 
-    validates :nsdname, presence: true, format: { with: Record::FQDN_REGEX, message: 'is not a fully qualified domain name' }
+    validates :nsdname, presence: true, format:
```
XaF: Same for `{`

lib/record_store/record/srv.rb
```diff
@@ -2,10 +2,26 @@ module RecordStore
   class Record::SRV < Record
     attr_accessor :priority, :port, :weight, :target
 
-    validates :priority, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
-    validates :port, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
-    validates :weight, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
-    validates :target, presence: true, format: { with: Record::CNAME_REGEX, message: 'is not a fully qualified domain name' }
+    validates :priority, presence: true, numericality:
```
XaF: Same for `{` for all of those following

lib/record_store/zone.rb
```diff
@@ -156,7 +160,8 @@ def validate_cname_records_for_same_fqdn
           when 'CNAME'
             errors.add(:records, "Multiple CNAME records are defined for #{record.fqdn}: #{record}")
           else
-            errors.add(:records, "A CNAME record is defined for #{cname_record.fqdn}, so this record is not allowed: #{record}")
+            errors.add(:records, "A CNAME record is defined for #{cname_record.fqdn}, "\
```
XaF: Either: store the error message in a tmp var, OR line returns for the call.
I think the first option would be cleaner

test/providers/ns1_test.rb
```diff
@@ -206,7 +206,8 @@ def test_update_changeset_for_fqdn_with_multiple_answers
 
       # Check
       updated_record_exists = @ns1.retrieve_current_records(zone: @zone_name).any? { |r| r == updated_record }
-      first_record_does_not_exist = @ns1.retrieve_current_records(zone: @zone_name).none? { |r| r == original_records[0] }
+      first_record_does_not_exist = @ns1.retrieve_current_records(zone: @zone_name).none? \
```
XaF: Instead:

```
      first_record_does_not_exist = @ns1.retrieve_current_records(zone: @zone_name)
        .none? { |r| r == original_records[0] }
```

lib/record_store/record/alias.rb
gooallen: Yeah, it complains with `Layout/IndentFirstHashElement`.

lib/record_store.rb
```diff
@@ -46,7 +46,10 @@ def secrets_path
     end
 
     def zones_path
-      @zones_path ||= Pathname.new(File.expand_path(config.fetch('zones_path'), File.dirname(config_path))).realpath.to_s
+      @zones_path ||= Pathname.new(
+        File.expand_path(config.fetch('zones_path'),
+        File.dirname(config_path)),
+        ).realpath.to_s
```
XaF: Isn't it going to complain about the fact that the closing parenthesis is not aligned with the `@zones_path` value?

gooallen: It's not complaining after aligned the bottom line with first line.
```
      @zones_path ||= Pathname.new(
        File.expand_path(config.fetch('zones_path'),
        File.dirname(config_path)),
      ).realpath.to_s
```

PR #145, opened 2019-10-22
bump up the version to 5.7.0 with OCI library

gooallen:
**What's the problem?**

OCI library as runtime dependency was missing 

**How does this fix it?**

Adding OCI library to the runtime dependency 

**How will I ship?**

- Merge this PR
- Have shipit it
- Release next version

**How to rollback?**

Revert the PR

### DIFF
--- a/record_store.gemspec
+++ b/record_store.gemspec
@@ -33,6 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'google-cloud-dns'
   spec.add_runtime_dependency 'ruby-limiter', '~> 1.0', '>= 1.0.1'
   spec.add_runtime_dependency 'ns1'
+  spec.add_runtime_dependency 'oci'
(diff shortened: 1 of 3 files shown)


record_store.gemspec
```diff
@@ -33,6 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'google-cloud-dns'
   spec.add_runtime_dependency 'ruby-limiter', '~> 1.0', '>= 1.0.1'
   spec.add_runtime_dependency 'ns1'
+  spec.add_runtime_dependency 'oci'
```
cttttt: 🤔 what if they release a version of the `oci` gem that has a completely different interface from the one that `record_store` depends on?  Consider a semver range, similar to the ones for dependencies like `fog`, `dnsimple` and `ruby-limiter`.

PR #146, opened 2019-11-06
Add API rate limit for NS1 provider and bump up the version to 5.7.1

gooallen:
**What's the problem?**
Make sure that the record_store for NS1 provider prevents throttling issue.

**How does this fix it?**
We might ask NS1 to increase API rate limit later, but currently, we have 180 limit per minute (10,800 per hour). How I implemented for the API rate limit is utilizing `Limiter` that is already being used for Dynect. 

I set `@queue = RateQueue.new(10000, interval: 3600)`. It will create a rate-limited queue which allows 10,000 operations per hour. (I can use up the rest of 800 if anyone wants)

Considering NS1 API rate limiting, it consumes differently depending on the API types. Users can call 10,800 queues for `GET` per hour. On the other hand, users only can use 3,600 queues for `POST` or `PUT` which are considered as expensive ones (three times expensive than `GET`) per hour.

- Simple `GET` will call single `@queue.shift`, such as `retrieve_current_records`, `zones`, and `records_for_zone` methods
- Expensive APIs other than `GET` will call `3.times { queue.shift }`, such as `add`, `update`, and `remove` methods

### Update Note
***2019-11-07***
No more fixed number of queues, it will be set up automatically based on `ratelimit_limit` & `ratelimit_period`. (I might be wrong, but I don't think I need `x-ratelimit-remaining` if I go with `RateQueue` with `limiter`). Currently, it's set `< x-ratelimit-limit: 900` and `< x-ratelimit-period: 300`. So, it will generate 3 queues per second. 

***But, I will definitely check `ns1-go` tomorrow morning how to leverage `x-ratelimit-remaining`***

***2019-11-08***
Updated the mechanism inspired by [ns1-go](https://github.com/terraform-providers/terraform-provider-ns1/blob/de56bbccd93a5484d01d9578a7ab03ebb372a628/vendor/gopkg.in/ns1/ns1-go.v2/rest/client.go#L250-L263). The rate limit will be determined by `x-ratelimit-remaining`, `x-ratelimit-period` from the API dynamically. It seems much stable and doesn't need to be hardcoded to configure the rate limit.

***2019-11-09***
Enhanced the API rate limit by patching `processing_response` to retrieve headers. While 'Monkey Patching', it was messed up entire functionalities so I refines the method and returns what it is supposed to return for previous design. (I think it's one of ways to reduce issues with 'Monkey Patch', called 'Refinements' with `refine processing_response do ... end`). As we discussed, release version will be `5.7.1`.

***2019-11-11**
Created a file for `patch` the API headers, named `patch_api_header` in `lib/record_store/provider/ns1` and replaced `refine NetHttp do` with `class NetHttp`.  Enhanced code based on comments.

**How to ship it?**
Merge & Have shipit

**How to roll-back?**
Revert


XaF:
> We might ask NS1 to increase API rate limit later, but currently, we have 180 limit per minute (10,800 per hour). How I implemented for the API rate limit is utilizing `Limiter` that is already being used for Dynect.
> 
> I set `@queue = RateQueue.new(10000, interval: 3600)`. It will create a rate-limited queue which allows 10,000 operations per hour. (I can use up the rest of 800 if anyone wants)

Why not using directly `RateQueue.new(180, interval: 60)` then, if that's the actual rate limit ?

> Considering NS1 API rate limiting, it consumes differently depending on the API types. Users can call 10,800 queues for `GET` per hour. On the other hand, users only can use 3,600 queues for `POST` or `PUT` which are considered as expensive ones (three times expensive than `GET`) per hour.
> 
> * Simple `GET` will call single `@queue.shift`, such as `retrieve_current_records`, `zones`, and `records_for_zone` methods
> * Expensive APIs other than `GET` will call `3.times { queue.shift }`, such as `add`, `update`, and `remove` methods

Do we know if using an operation in the rate limit of the 'expensive' ones really impacts the 'unexpensive' ones? (i.e. if I run out of my rate limit for the expensive operations, can I still 'GET' or will it also be 'blocked' at this point ?)
1. If it is, then we might need to figure out exactly how they compute this on the NS1 side
2. If it isn't... why doing that whole complicated setup instead of just using two separate queues depending on the rate limit we have? You setup a `@write_queue` and a `@read_queue` (as it seems that expensive = write operations, unexpensive = read operations) with different values, and you shift from those depending on the operation you are running.

gooallen:
@XaF 
> Why not using directly RateQueue.new(180, interval: 60) then, if that's the actual rate limit ?

I think that's better. I think I was sort of biased to use `interval: 3600` from README. 

> Do we know if using an operation in the rate limit of the 'expensive' ones really impacts the 'unexpensive' ones? 

I think I need to be clear here. Here's my understanding from [NS1 website](https://help.ns1.com/hc/en-us/articles/360020250573-About-API-rate-limiting). For example, 180 queues are generated per minute. Users can call any regardless of 'expensive' ones or 'inexpensive' ones. The only difference is the 'expensive' ones will take 3 tokens and 'inexpensive' ones will take 1 token. So remaining token only matters. 

> (i.e. if I run out of my rate limit for the expensive operations, can I still 'GET' or will it also be 'blocked' at this point ?)

You can still 'GET' if users haven't used remaining tokens up. Let's say users call 30 'POST's in a minute. There are still available 90 tokens before next minute comes.

XaF:
> I think I need to be clear here. Here's my understanding from [NS1 website](https://help.ns1.com/hc/en-us/articles/360020250573-About-API-rate-limiting). For example, 180 queues are generated per minute. Users can call any regardless of 'expensive' ones or 'inexpensive' ones. The only difference is the 'expensive' ones will take 3 tokens and 'inexpensive' ones will take 1 token. So remaining token only matters.

It seems to be your interpretation of it, yes, but I can't read that from the link you've given me (which does not mean you're not right, though!). The example they give is '900 tokens for 300 seconds' for the inexpensive operations and '300 tokens for 300 seconds' for the expensive ones, which gives your 1 for 3 tokens approach. But in their explanation, they always talk about grabbing 1 token from the available ones, and never say that the token pools are coming from the same place, instead of having one pool for inexpensive and one pool for expensive. Now, that might be the case, but I'm really unclear about this from the source link, which is why I suggested checking with them :)

> You can still 'GET' if users haven't used remaining tokens up. Let's say users call 30 'POST's in a minute. There are still available 90 tokens before next minute comes.

You're saying that in the case where I haven't used all my tokens. My question was if we already used all the tokens possible with POST operations, did it deplete the GET tokens too?
Let's say I have 180 tokens available and I do 60 POST requests. Then I try to do a GET, what happens? (this would be a way to check the theory without asking them, but I would prefer asking them :))

gooallen:
>It seems to be your interpretation of it, yes, but I can't read that from the link you've given me (which does not mean you're not right, though!). The example they give is '900 tokens for 300 seconds' for the inexpensive operations and '300 tokens for 300 seconds' for the expensive ones, which gives your 1 for 3 tokens approach. But in their explanation, they always talk about grabbing 1 token from the available ones, and never say that the token pools are coming from the same place, instead of having one pool for inexpensive and one pool for expensive. Now, that might be the case, but I'm really unclear about this from the source link, which is why I suggested checking with them :)

Yeah, maybe it is. The best way is that I confirm with them :))
Thanks for the feedback! 👍 
Will share what I get from NS1 with you!

lib/record_store/provider/ns1.rb
```diff
@@ -1,10 +1,12 @@
 require_relative 'ns1/client'
+require 'net/http'
 
 module RecordStore
   class Provider::NS1 < Provider
     class Error < StandardError; end
 
     class << self
+      include Limiter
```
XaF: Does not seem needed anymore

lib/record_store/version.rb
```diff
@@ -1,3 +1,3 @@
 module RecordStore
-  VERSION = '5.7.0'.freeze
+  VERSION = '5.8.0'.freeze
```
XaF: Would probably go for `5.7.1` instead as this is not really a new feature if we don't change behavior nor break it.
It's just transparent for the user, so a 'patch' makes sense to me here (we're patching the behavior to fit with what should be done and avoid rate limits). Maybe @cttttt will have a different feeling though?

lib/record_store/provider/ns1.rb
gooallen: Oops, nice catch! interesting that Rubocop didn't catch it.

lib/record_store/version.rb
gooallen: I see. Yeah I was debating between 'new feature' and 'patch'. I decided to go with 'feature' because I believe it change bahavior (now it controls the rate limit by 'sleep'). But I also get your point. I want to wait for @cttttt 's thought as well.

lib/record_store/provider/ns1.rb
```diff
@@ -219,6 +227,23 @@ def symbolize_keys(hash)
       def secrets
         super.fetch('ns1')
       end
+
+      def api_rate_limit
+        uri = URI('https://api.nsone.net/v1/monitoring/regions')
```
XaF: Why are we doing that extra call before any other API call? That's not necessarily the endpoint all the other calls are going to be using. IIRC, in `ns1-go`, they were using the headers of the call they just made, weren't they?
In this case, you are:
1/ duplicating every API call by adding that GET call (hence actively reducing the rate at which we can effectively use the API)
2/ considering only the headers giving us the token information for the GET call (as discussed with NS1: GET, PUT, DELETE and POST have all different token counters, so here we would only know about the GET tokens no matter what method the call used).

I realize that `client.record` might not return the headers at this point, which means we might need to patch the method that actually does the API call to either return the headers or implement the rate limiting approach!

lib/record_store/version.rb
XaF: My thought generally is:
- Does it remove any existing function or change the way to call any existing function ? => major
- Does it add a new function or new parameters to an existing function without breaking its current behavior for users ? => minor
- Else => patch (basically, does it do anything in the background that the user won't be impacted by other than positively, like fixing bugs or adding a process to avoid bugs)

But I'm not really caring about the version number. I just prefer to avoid reaching version 5.567.0 too soon :p

lib/record_store/provider/ns1.rb
gooallen: It might be dumb question but, aren't they duplicating calls anyways if I call them from other URIs?
- `https://api.nsone.net` for zones and records
- `https://api.nsone.net/v1/monitoring/regions` for headers (nothing related to zones and records)

lib/record_store/version.rb
cttttt: As long as this is well tested (and practically, as long as it touches this provider only (it hasn't really seen prime time)) a patch release sounds good 👍🏽.

@XaF 's impact `switch` is pretty spot on.  This update isn't intended to change behaviour so it should be okay if folks automatically pick it up.  See https://semver.org/ for the canonical meanings of major, minor, and patch.

lib/record_store/provider/ns1.rb
gooallen: Third option might take longer than expected. I just checked the closed and opened PR in the repo. Last closed PR was merged Dec, 2018, opened PR is still waiting for approval from Jul, 2019.

gooallen: I will probably go with ns1-ruby patch.
>lib/ns1/api/regions.rb
```ruby
module NS1
  module API
    module Regions
      #
      # Returns the list of all monitoring regions for the account
      #
      # @return [NS1::Response]
      #
      def regions()
        perform_request(HTTP_GET, "/v1/monitoring/regions")
      end
    end
  end
end
```
>lib/ns1/api.rb
```ruby
    def self.included(base)
      base.send :include,
        NS1::API::Zones,
        NS1::API::Records
        NS1::API::Records,
        NS1::API::Regions
    end
```
Something like that..

cttttt: It's unfortunate that [only the body is captured](https://github.com/kitop/ns1-ruby/blob/2635616453b37b4ab7e417c08f401ec4e694622c/lib/ns1/transport/net_http.rb#L28-L38) in `ns1` 😢 

I agree with @XaF that it's weird that we're making calls outside of the `ns1` gem...and it's also weird that we're checking the headers on a request that's unrelated to the request we're making (this will result in our _usually_ halving the number of tickets we have, but _usually_...we've established that there are different buckets of tickets for POST, GET, etc).

He's spot on on options here.

lib/record_store/provider/ns1.rb
```diff
@@ -219,6 +227,23 @@ def symbolize_keys(hash)
...
+          http.request(req)
+        end
+
+        ratelimit_remaining = res.to_hash["x-ratelimit-remaining"].first.to_i
+        ratelimit_period = res.to_hash["x-ratelimit-period"].first.to_i
+
+        if ratelimit_remaining < 2
+          sleep(ratelimit_period)
+        end
+        sleep(ratelimit_period / ratelimit_remaining)
```
cttttt: 🤔 on line 242, if we have a low number of tokens available, we'll sleep for the entire ratelimit_period (fair).  Do you also intend to sleep for possibly the entire rate limit?   Feels like lines 243 and 245 shouldn't both run, but lemme know what ur intentions are.

XaF: Good catch! In [`ns1-go`](https://github.com/ns1/ns1-go/blob/da571caa8ebc28580715f677f97799aefde48679/rest/client.go#L290-L295) they use a separate method that "returns" to ensure we're only doing one of the cases. Here, a `else` would probably be sufficient.

lib/record_store/provider/ns1.rb
XaF: From @cttttt 's comment and your analysis of their merging of the requests, patching (from this project) the `NS1::Transport::process_response` method might be a good in-between for this, as this methods has the `response` and thus the headers! (this is option 1 above) This would be patching the client even though we're not having the upstream client profit from that patch!
Option 2 of writing fully our own client could work too, but I feel like it would take longer. Up to you to decide at this point @gooallen !

lib/record_store/provider/ns1.rb
gooallen: Right, that's right. I mixed a bit with Go. As Raphaël mentioned, `else` should be sufficient.

lib/record_store/version.rb
gooallen: Cool, now it's solid that we are going with patch, `5.7.1`.

lib/record_store/provider/ns1.rb
gooallen: Yeah, it's very unfortunate.. :( 
Seems like the option 1 is the best choice for now than the option 2 taking longer. Let me check `NS1::Transport::process_response` method as @XaF suggested to take a look. Let me see how it goes. Thanks, both!

gooallen: It seems possible to call headers from ns1-ruby. Hopefully, it works out in the way we want :D

lib/record_store/provider/ns1.rb
```diff
@@ -219,6 +220,35 @@ def symbolize_keys(hash)
       def secrets
         super.fetch('ns1')
       end
+
+      # Patch the method which retrieves headers for API rate limit dynamically
+      module NS1::Transport
```
XaF: Is this working from here? Reading the code as-is I would understand that you are creating a module `RecordStore::Provider::NS1::NS1::Transport` which does not seem quite what we would want.

What I would suggest is to put this in a separate `ns1_client_patch.rb` (or name of your choosing!) file

lib/record_store/provider/ns1.rb
```diff
@@ -219,6 +220,35 @@ def symbolize_keys(hash)
...
+        refine NetHttp do
+          def process_response(response)
+            ratelimit = ['remaining', 'period']
+              .map { |k| [k, response["x-ratelimit-#{k}"].to_i] }.to_h
+
+            ratelimit_remaining = ratelimit["remaining"]
```
XaF: Using the map _then_ doing this seems overkill. Just pull the data once directly from the headers!

lib/record_store/provider/ns1.rb
```diff
@@ -219,6 +220,35 @@ def symbolize_keys(hash)
...
+            ratelimit_period = ratelimit["period"]
+
+            if ratelimit_remaining < 2
```
XaF: You can probably replace this whole block by:

```
sleep(ratelimit_period / max(1, ratelimit_remaining))
```
Also, not sure about how ruby handles divisions with integers? Might want to convert those to floats?

lib/record_store/provider/ns1.rb
```diff
@@ -1,4 +1,5 @@
 require_relative 'ns1/client'
+require 'net/http'
```
cttttt: Reminder: Move this where the refinement / monkey patch goes.

gooallen: Right, this doesn't need to be here. Thanks!

lib/record_store/provider/ns1.rb
gooallen: I think just the snippet that you provided is enough. I tested with float type and it slows performance down with when it returns 0.2, 0.3 (integer will be 0, so no waiting time for those).

lib/record_store/provider/ns1.rb
gooallen: Now it's `patch_api_header.rb` in `lib/record_store/provider/ns1/`. :)

lib/record_store/provider/ns1/patch_api_header.rb
```diff
@@ -0,0 +1,21 @@
+require 'net/http'
+
+# Patch the method which retrieves headers for API rate limit dynamically
+module NS1::Transport
+  class NetHttp
+    def process_response(response)
+      sleep(response.to_hash["x-ratelimit-period"].first.to_i /
```
cttttt: Have you confirmed that this prevents the lack of precision in this division?  Consider looking at the implementation of `Integer./` to add context to @XaF 's comment.

gooallen: I have confirmed that this prevents the lack of precision in this devision. I actually left more context in @xaf's  comment but I decided not to.

Before I put, `to_f`. 300 (period) / 899 (remaining) was 0.
Now it returns 0.33..... 

But if you want me to handle in case headers aren't valid integer or headers aren't integer, I can consider looking at the `Integer.`.

cttttt: Gotcha 👍 .  Had a bit of a brain freeze there.

PR #150, opened 2019-12-05
NS1: Escape colons on download, unescape colons on upload in TXT rdata

cttttt:
**What's the problem**

- If a `record_store` YML file contains text resource records with rrdata containing semi-colons, when these records are pushed to DNS providers:
  - Those colons (`;`) are POSTed to NS1 as escaped colons (`\;`).  This turns out to not be a problem.
  - **Problem**: During a subsequent diff, those records are considered to be out of sync, because of the difference in escaping.

**Why'd's it happen?**

The NS1 provider introduces escaping when loading records, but doesn't strip the escaping when records are POSTed.

**Why do providers escape semi-colons on GET?**

- Because the API of one of the providers, Dynect, returns records with escaped semi-colons.  All other providers seem to be following this pattern to provide the rest of `record_store` normalized data.
- Just in case anyone's storing those escaped records in marshalled Record `yml` files from when Dynect was the only show in town.  By ensuring that all semi-colons are escaped on GET, the internal representation of Records should be consistent across all providers, as well as from those YAML files.

**How does this fix the problem?**

- By simply un-escaping the semi-colons when records are written out to NS1.

### DIFF
--- a/lib/record_store/provider/ns1.rb
+++ b/lib/record_store/provider/ns1.rb
@@ -245,7 +245,7 @@ module RecordStore
         if record.is_a?(Record::MX)
           [record.preference, record.exchange]
         elsif record.is_a?(Record::TXT) || record.is_a?(Record::SPF)
-          [Record.long_quote(record.txtdata)]
+          [Record.long_quote(record.txtdata).gsub('\;', ';')]
         elsif record.is_a?(Record::CAA)
           [record.flags, record.tag, record.value]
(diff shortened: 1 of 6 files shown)


XaF:
This makes sense. Note that I would vote for inverting this behavior in all providers if that's something we can do someday...

cttttt:
Agreed.  This feels like a pitfall.

PR #155, opened 2020-03-31
Introduce sshfp resource record

ryandv:
# Context

record_store does not model `SSHFP` RRs internally, and so will be unable to both export such records from DynECT and apply such records to DNSimple.

Note that [NS1 does not support the SSHFP RR type](https://help.ns1.com/hc/en-us/articles/360022091074-NS1-supported-DNS-record-types); any zones that must serve such records must therefore be exclusively managed by DNSimple.

# Solution
This PR introduces a `Record::SSHFP` class and adds clauses to handle reading/writing SSHFP records from/to DNSimple.


XaF:
> Note that NS1 does not support the SSHFP RR type; any zones that must serve such records must therefore be exclusively managed by DNSimple.

Is that something that will be checked and raise an error during the validation step ?

ryandv:
NS1 will actually return intelligible error messages (`SSHFP is not a supported record type in NS1`) when trying to write these RRs via their API.

I've written some tests to capture this behaviour, and also surfaced the validation error messages from the CLI for better diagnostics.

XaF:
> NS1 will actually return intelligible error messages (`SSHFP is not a supported record type in NS1`) when trying to write these RRs via their API.
> 
> I've written some tests to capture this behaviour, and also surfaced the validation error messages from the CLI for better diagnostics.

The main problem is that once we pass the validation step, any issue while deploying something with the `record_store` will leave the state in inconsistency, requiring potential manual intervention. Which is why catching it before making any change might be a better approach here!

lib/record_store/cli.rb
```diff
@@ -104,7 +104,7 @@ def apply
       end
 
       zones.each do |zone|
-        abort("Attempted to apply invalid zone: #{zone.name}") unless zone.valid?
+        abort("Attempted to apply invalid zone: #{zone.name}: #{zone.errors.full_messages.join(',')}") unless zone.valid?
```
cttttt, line 107: Perhaps this belongs in its own PR as it's not directly related to adding a new record type.

lib/record_store/record/sshfp.rb
```diff
@@ -0,0 +1,56 @@
+module RecordStore
+  class Record::SSHFP < Record
+    attr_accessor :algorithm, :fptype, :fingerprint
+
+    # https://www.iana.org/assignments/dns-sshfp-rr-parameters/dns-sshfp-rr-parameters.xhtml
```
cttttt, line 5: 👍🏽✨

lib/record_store/record/sshfp.rb
```diff
@@ -0,0 +1,56 @@
...
+      ED448 = 6
+    end
+
+    module FingerprintTypes
+      RESERVED = 0
+      SHA_1 = 1
+      SHA_256 = 2
+    end
+
+    FINGERPRINT_REGEX = /\A[[:xdigit:]]+\Z/
```
cttttt, line 22: You probably want to match against the entirety of the string.

```suggestion
    FINGERPRINT_REGEX = /\A[[:xdigit:]]+\z/
```

lib/record_store/cli.rb
```diff
@@ -103,9 +103,11 @@ def apply
         exit
       end
 
-      zones.each do |zone|
-        abort("Attempted to apply invalid zone: #{zone.name}: #{zone.errors.full_messages.join(',')}") unless zone.valid?
+      if (invalid_zone = zones.find(&:invalid?))
+        abort("Attempted to apply invalid zone: #{invalid_zone.name}: #{invalid_zone.errors.full_messages.join(',')}")
+      end
```
XaF, line 108: One thing that would be nice to avoid back-and-forth here would be to go through that loop, list all invalid zones, and then abort with one message listing all the invalid zones right away?
That way, if I make two mistakes, I have the two errors right away, instead of needing to fix one to see the other.

XaF:
> A few very small comments. Re: Validation, see [`supports_alias?`](https://github.com/Shopify/record_store/blob/335cfb416e13e1deb92576c0742ad35236120093/lib/record_store/provider.rb#L44-L46) for tips on how another partially supported record is handled.

Would totally make sense to have a `supports_sshfp?` method added, including the parameter for the zone.

lib/record_store/cli.rb
ryandv: Ok - as per @cttttt's [comment](https://github.com/Shopify/record_store/pull/155#discussion_r401693619) I'll revert out all changes to cli.rb & make a follow-up PR with your suggestion to list all invalid zones =)

.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml
```diff
@@ -872,9 +872,6 @@ Lint/EmptyEnsure:
 Lint/EmptyInterpolation:
   Enabled: true
 
-Lint/EndInMethod:
```
ryandv, line 875: Rubocop was updated to 0.81.0 in CI (the gemspec does not constrain versions). This cop has been renamed to the (already-enabled) `Style/EndBlock` cop.

lib/record_store/provider.rb
```diff
@@ -54,7 +54,7 @@ def build_zone(zone_name:, config:)
       end
 
       # returns an array of Record objects that match the records which exist in the provider
-      def retrieve_current_records(zone:, stdout: $stdout) # rubocop:disable Lint/UnusedMethodArgument
```
ryandv, line 57: rubocop-0.81.0 finds these cops redundant now.

PR #164, opened 2020-04-16
Add PTR record support

ryandv:
Add support for reading/writing PTR records to both NS1 and DNSimple. Such records can already be read using the DynECT provider.


test/record_test.rb
```diff
@@ -365,4 +371,20 @@ def test_spf_is_deprecated
     Record::SPF.new(fqdn: 'dns-test.shopify.io.', txtdata: 'v=spf1 -all', ttl: 3600)
     assert_includes($stderr.string, "SPF record type is deprecated (See RFC 7208 Section 14.1)")
   end
+
+  def test_valid_ptr_record
+    assert_predicate(Record::PTR.new(
+      fqdn: '4.0.41.198.in-addr.arpa',
```
sbfaulkner, line 377: I don't know the spec for PTR offhand, but should we validate that it's only up to (or maybe exactly - really not sure) four octets, and that they are numbers in the range 0-255?

test/providers/dnsimple_test.rb
```diff
@@ -328,6 +328,21 @@ def test_apply_sshfp_changeset
...
+          record.fqdn == sshfp_record.fqdn
+      end
+
+      assert_equal(1, matching_records.size, 'could not find the SSHFP record that was just created')
+      assert_equal(Record::SSHFP::Algorithms::ED25519, matching_records.first.algorithm)
+      assert_equal(Record::SSHFP::FingerprintTypes::SHA_256, matching_records.first.fptype)
+      assert_equal(
+        '4e0ebbeac8d2e4e73af888b20e2243e5a2a08bad6476c832c985e54b21eff4a3',
+        matching_records.first.fingerprint
+      )
```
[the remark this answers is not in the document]
ryandv: Yes, it's unrelated, but deliberately so. I thought I could sneak in some changes here to strengthen tests from the [SSHFP pull request](https://github.com/Shopify/record_store/pull/164) as I noticed they weren't really asserting that any records had been persisted.

I'm willing to just back this change out. The response is stored on `cassette` anyway so it's really only verifying a change took place on the remote server the very first time it's run.

test/record_test.rb
ryandv: Had to take a second look over [RFC 1035 § 3.5](https://tools.ietf.org/html/rfc1035):

> Domain names in the IN-ADDR.ARPA domain are defined to have up to four labels in addition to the IN-ADDR.ARPA suffix.  Each label represents one octet of an Internet address, and is expressed as a character string for a decimal value in the range 0-255 (with leading zeros omitted except in the case of a zero octet which is represented by a single zero).

I can write up a validation for this.

ryandv:
With respect to adding support in the `GoogleCloudDNS` and `OracleCloudDNS` providers, I would prefer to follow up with a separate PR for now, unless there are strong objections.

Providers must explicitly declare what record types they support. I have characterized the behaviour in some [new tests](https://github.com/Shopify/record_store/pull/164/files#diff-464a1d0749bba78fcf58be3ca5008e98). Otherwise the Zone will raise a "PTR is not a supported record type" validation error.

test/providers/dnsimple_test.rb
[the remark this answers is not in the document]
sbfaulkner: I'm fine with the occasional drive-by  :-)

mostly making sure I wasn't missing something, and can see the commit now

for future, maybe the commit comment could be more description for cases like this, or a note added to the PR description to call out the out-of-scope change -- ie. the comment "Assorted test cleanup" didn't clearly tell me what it was doing, and if I were looking at this and doing some archaeology in future it might be confusing

lib/record_store/provider/dnsimple.rb
```diff
@@ -4,7 +4,7 @@ module RecordStore
   class Provider::DNSimple < Provider
     class << self
       def record_types
-        super | Set.new(['SSHFP'])
+        super | Set.new(['PTR', 'SSHFP'])
```
sbfaulkner, line 7: syntactic preference on my part to use `%w(PTR SSHFP)` for things like this... I can see how this could be inconsistent with other code in this repo

ryandv: Switched over to the alternate array literal syntax =)

ryandv:
Just pushed up some additional validations for PTR FQDNs:

* Must specify 1-4 octets
* Octets must be within range 0-255
* FQDN must be within the `in-addr.arpa` zone

PR #165, opened 2020-04-23
add option to include ignored records in diff

sbfaulkner:
similar to the addition of a `--all` option for the `list` command to include ignored records,
this PR adds support for `--all` to the `diff` command as well, to show differences between the ignored records in the zone definition, and the actual records in the provider(s)

it also makes it possible to specify which zones to diff (rather than all defined ones)

sample output...
```
$ bin/record-store diff --all kafka.shopifysvc.com
Diffing 1 zone(s)

Zone: kafka.shopifysvc.com
--------------------
Provider: DNSimple
Add:
 - [NSRecord] kafka.shopifysvc.com. 3600 IN NS dns1.p07.nsone.net.
 - [NSRecord] kafka.shopifysvc.com. 3600 IN NS dns2.p07.nsone.net.
 - [NSRecord] kafka.shopifysvc.com. 3600 IN NS dns3.p07.nsone.net.
 - [NSRecord] kafka.shopifysvc.com. 3600 IN NS dns4.p07.nsone.net.
====================
```

### DIFF
--- a/test/changeset_test.rb
+++ b/test/changeset_test.rb
@@ -189,7 +189,7 @@ class ChangesetTest < Minitest::Test
       name: 'dns-test.shopify.io',
       config: {
         providers: ['DynECT'],
-        ignore_patterns: [{ type: 'NS', fqdn: 'dns-test.shopify.io.'}],
+        ignore_patterns: [{ type: 'NS', fqdn: 'dns-test.shopify.io.' }],
       },
       records: [{
         type: 'NS',
@@ -234,7 +234,7 @@ class ChangesetTest < Minitest::Test
       name: 'dns-test.shopify.io',
       config: {
         providers: ['DynECT'],
-        ignore_patterns: [{ type: 'NS', fqdn: 'dns-test.shopify.io.'}],
+        ignore_patterns: [{ type: 'NS', fqdn: 'dns-test.shopify.io.' }],
       },
       records: [{
         type: 'NS',
--- a/test/changeset_test.rb
+++ b/test/changeset_test.rb
@@ -184,6 +184,97 @@ class ChangesetTest < Minitest::Test
     end
   end
 
+  def test_changeset_build_from_creates_changeset_excluding_ignored_records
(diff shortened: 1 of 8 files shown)


dev.yml
```diff
@@ -2,7 +2,7 @@
 name: recordstore
 
 up:
-  - ruby: 2.5.0
+  - ruby: 2.5.7
```
ryandv, line 5: Just curious - what motivated this ruby upgrade?

test/changeset_test.rb
```diff
@@ -184,6 +184,97 @@ def test_changeset_build_from_creates_changeset_from_diff_between_zone_and_provi
...
+      }, {
+        type: 'ALIAS',
+        ttl: 60,
+        fqdn: 'dns-test.shopify.io',
+        alias: 'dns-test.herokuapp.com.',
+      }]
+    )
+
+    # Cassette matches zone's records except with an additional NS record
+    VCR.use_cassette('dynect_retrieve_current_records') do
```
ryandv, line 223: Are these fixtures checked into version control? CI appears to have trouble finding the appropriate cassettes.

dev.yml
sbfaulkner: was just testing this to see if it's somehow related to the failing test(s).
they don't fail locally :-(

test/changeset_test.rb
sbfaulkner: I didn't think it was the same cassette as my test that was not working in CI -- I used a pre-existing cassette

dev.yml
sbfaulkner: reverted this ruby version bump

the issue was the api host name changed with a google-cloud-dns gem version

(I've updated cassettes and locked the gem version)

test/changeset_test.rb
sbfaulkner: the issue was the api host name changed with a google-cloud-dns gem version

(I've updated cassettes and locked the gem version)

PR #167, opened 2020-05-12
Add rate limiting for DNSimple provider

mathu97:
Patches the DNSimple client to add API rate limiting. This follows a similar approach to how it was done for the [NS1 provider](https://github.com/Shopify/record_store/blob/master/lib/record_store/provider/ns1/patch_api_header.rb#L6-L8).

### DIFF
--- a/lib/record_store/provider/dnsimple/patch_api_header.rb
+++ b/lib/record_store/provider/dnsimple/patch_api_header.rb
@@ -1,0 +1,30 @@
+# Patch Dnsimple client method which retrieves headers for API rate limit dynamically
+module Dnsimple
+  class Client
+    def execute(method, path, data = nil, options = {})
+      response = request(method, path, data, options)
+      rate_limit_sleep(response.headers["x-ratelimit-reset"].to_i, response.headers["x-ratelimit-remaining"].to_i)
+
+      case response.code
+      when 200..299
+        response
+      when 401
+        raise AuthenticationFailed, response["message"]
+      when 404
+        raise NotFoundError, response
+      else
+        raise RequestError, response
+      end
+    end
+
+    private
+
+    def rate_limit_sleep(rate_limit_reset, rate_limit_remaining)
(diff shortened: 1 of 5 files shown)


XaF:
That looks good! 👍 
One missing thing is to bump the version in `lib/record_store/version.rb`

lib/record_store/provider/dnsimple/patch_api_header.rb
```diff
@@ -0,0 +1,26 @@
...
+        raise NotFoundError, response
+      else
+        raise RequestError, response
+      end
+    end
+
+    private
+
+    def rate_limit_sleep(rate_limit_reset, rate_limit_remaining)
+      sleep([0, rate_limit_reset - Time.now.to_i].max / (rate_limit_remaining + 1).to_f)
```
XaF, line 23: Wondering if we could avoid to call `sleep` when the result is `0`
Something like:

```suggestion
wait_time = [0, rate_limit_reset - Time.now.to_i].max / (rate_limit_remaining + 1).to_f

sleep(wait_time) if wait_time > 0
```

Would of course require you to adapt the tests below to make sure that `sleep` is **not called* when the `wait_time` is computed to be `0`

mathu97: Gotcha, that makes sense to do. I'll add in this change.

XaF: Another helpful thing (at least for the tests) would be to return a boolean depending on what that method does:

```
return false unless wait_time > 0
sleep(wait_time)
return true
```

Which would allow to directly check that the method return value is `false` when you expect not to have to wait

mathu97: Ah ya after looking a bit more into this, it seems that `MiniTest` doesn't exactly have something to test if a function "is not called". Some ideas were to assert an error and check if error is raised, similar to checking for bool as you mentioned, but that would mean having to catch the error whenever `rate_limit_sleep` is called. I'll try out returning a bool.

PR #168, opened 2020-05-20
handle non-NS records returned as authority

sbfaulkner:
The original implementation assumed all records in the authority section were all NS records, but other record types can be present. This PR updates to filter out non-NS records.


cttttt:
🎗- bump the version

sbfaulkner:
@cttttt 
see also...
#170 -- which this branch is based on
and
#171 -- which does a version bump

PR #169, opened 2020-05-20
Validate authority for select zones

sbfaulkner:
this PR is a small enhancement for `record-store validate_authority …`

1. sorts zones for more predictable zone processing
2. supports passing zone names to process on the command line (instead of always processing all zones)


sbfaulkner:
added tests fwiw

PR #170, opened 2020-05-20
Handle unreachable host when looking up authority

sbfaulkner:
this PR implements retry if an authoritative NS is unreachable

#### update: test added


XaF:
Don't forget the version increase though

PR #172, opened 2020-06-10
Add print when waiting for rate-limit to elapse so deployment doesn't timeout

mathu97:
This PR is to prevent Shipit from timing out because of a lack of output when waiting for the NS1 rate-limit to elapse. To do this we emit a "waiting" message `Waiting on NS1 provider` every 10 seconds to let Shipit know that we're still alive and just waiting for our rate-limit to elapse.

### DIFF
--- a/lib/record_store/provider/dnsimple/patch_api_header.rb
+++ b/lib/record_store/provider/dnsimple/patch_api_header.rb
@@ -1,15 +1,17 @@
+require_relative '../provider_utils/rate_limit'
+
 # Patch Dnsimple client method which retrieves headers for API rate limit dynamically
 module Dnsimple
   class Client
     def execute(method, path, data = nil, options = {})
       response = request(method, path, data, options)
-      rate_limit_sleep(response.headers["x-ratelimit-reset"].to_i, response.headers["x-ratelimit-remaining"].to_i)
+      rate_limit_sleep(response.headers['x-ratelimit-reset'].to_i, response.headers['x-ratelimit-remaining'].to_i)
 
       case response.code
       when 200..299
         response
       when 401
-        raise AuthenticationFailed, response["message"]
+        raise AuthenticationFailed, response['message']
       when 404
         raise NotFoundError, response
       else
@@ -22,9 +24,10 @@ module Dnsimple
     def rate_limit_sleep(rate_limit_reset, rate_limit_remaining)
       rate_limit_reset_in = [0, rate_limit_reset - Time.now.to_i].max
       rate_limit_periods = rate_limit_remaining + 1
-      wait_time = rate_limit_reset_in / rate_limit_periods.to_f
+      sleep_time = rate_limit_reset_in / rate_limit_periods.to_f
 
-      sleep(wait_time) if wait_time > 0
+      rate_limit = RateLimit.new('DNSimple')
+      rate_limit.sleep_for(sleep_time)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


lib/record_store/provider/dnsimple/patch_api_header.rb
```diff
@@ -19,12 +19,17 @@ def execute(method, path, data = nil, options = {})
...
 
-    def rate_limit_sleep(rate_limit_reset, rate_limit_remaining)
+    def rate_limit_sleep(_rate_limit_reset, _rate_limit_remaining)
       rate_limit_reset_in = [0, rate_limit_reset - Time.now.to_i].max
       rate_limit_periods = rate_limit_remaining + 1
-      wait_time = rate_limit_reset_in / rate_limit_periods.to_f
+      sleep_time = rate_limit_reset_in / rate_limit_periods.to_f
 
-      sleep(wait_time) if wait_time > 0
+      while sleep_time > 0
```
ryandv, line 27: Looks like this may be duplicated with the same loop in the NS1 patch.

Is it possible to extract this into a Module, or a Class?

lib/record_store/provider/dnsimple/patch_api_header.rb
```diff
@@ -19,12 +19,17 @@ def execute(method, path, data = nil, options = {})
 
     private
 
-    def rate_limit_sleep(rate_limit_reset, rate_limit_remaining)
+    def rate_limit_sleep(_rate_limit_reset, _rate_limit_remaining)
```
ryandv, line 22: As an optional cleanup: Since the method is private to this class, it should be possible to change this method signature to get rid of the unused parameters - there won't be many callsites to change (the callsite in `#execute`, as well as some tests in `test/providers/dnsimple_test.rb`.

lib/record_store/provider/ns1/patch_api_header.rb
```diff
@@ -4,8 +4,16 @@
 module NS1::Transport
   class NetHttp
     def process_response(response)
-      sleep(response.to_hash["x-ratelimit-period"].first.to_i /
-        [1, response.to_hash["x-ratelimit-remaining"].first.to_i].max.to_f)
+      # Printing something to avoid shipit from timing out if rate limit sleep is long
```
ryandv, line 7:
```suggestion
```

lib/record_store/provider/dnsimple/patch_api_header.rb
XaF, line 22: Why change the variable names ? Seems those are not used anymore here ?
Where does the value `rate_limit_reset` come from in the next line ?

mathu97: oops i think this might have been a rubocop thing, have it set to format on save, and was playing around with the rubocop configs haha. I'll fix this.

lib/record_store/provider/dnsimple/patch_api_header.rb
mathu97: Should this be put into some separate `utils` folder in `record_store/provider/`?

ryandv: Could be =) Looks like functionality that would want to be shared across all providers, so that makes sense.

Consider also dumping it directly into `record_store/provider` if you don't want to create the folder/submodule. Just presenting alternatives.

Either way I trust your judgment on this one.

PR #173, opened 2020-08-14
Retry read operations on connection errors

cttttt:
**What's the problem?**

- `record_store` doesn't really protect its transactions.  A failure during during a CI run in which `record_store` is run could result in inconsistencies between a `record_store` configuration and DNS providers.
- For this reason, it's important that `record_store` does not fail unnecessarily.
- Transient network problems (connection timeouts, connection resets), although rare, could cause `record_store` to fail, but can be retried if they occur during read operations.

**How does this help?**

By performing a limited number of retries where a request hits connection timeouts or connection resets.  The threshold of errors is reached when 5 connection timeouts or 5 connection resets are encountered on a single connection attempt.  These counters could be expanded as new network issues emerge in provider APIs.

**Review tips?**

I wasn't able to figure out a way to perform tests directly using VCR, so I fell back to Webmock.  If it's possible to simulate a connection reset or timeout in VCR, please let me know.

### DIFF
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -142,9 +142,13 @@ module RecordStore
           rescue Net::OpenTimeout
             raise if max_timeouts <= 0
             max_timeouts -= 1
+
+            $stderr.puts("Retrying after a connection timeout")
           rescue Errno::ECONNRESET
             raise if max_conn_resets <= 0
             max_conn_resets -= 1
+
+            $stderr.puts("Retrying in #{delay}s after a connection reset")
             backoff_sleep(delay)
             delay = [delay * backoff_multiplier, max_backoff].min
           end
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -129,7 +129,13 @@ module RecordStore
         dns.getresource(zone_name, Resolv::DNS::Resource::IN::SOA).mname.to_s
       end
 
-      def retry_on_connection_errors(max_timeouts: 5, max_conn_resets: 5)
+      def retry_on_connection_errors(
+        max_timeouts: 5,
+        max_conn_resets: 5,
+        delay: 1,
+        backoff_multiplier: 2,
+        max_backoff: 10
+      )
         loop do
           begin
             return yield
@@ -139,8 +145,14 @@ module RecordStore
           rescue Errno::ECONNRESET
             raise if max_conn_resets <= 0
             max_conn_resets -= 1
+            backoff_sleep(delay)
+            delay = [delay * backoff_multiplier, max_backoff].min
           end
         end
+      end
+
+      def backoff_sleep(delay)
+        sleep(delay)
       end
     end
   end
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -128,6 +128,20 @@ module RecordStore
 
         dns.getresource(zone_name, Resolv::DNS::Resource::IN::SOA).mname.to_s
       end
+
+      def retry_on_connection_errors(max_timeouts: 5, max_conn_resets: 5)
+        loop do
+          begin
+            return yield
+          rescue Net::OpenTimeout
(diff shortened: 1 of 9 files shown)


lib/record_store/provider.rb
```diff
@@ -128,6 +128,20 @@ def master_nameserver_for(zone_name)
...
+      def retry_on_connection_errors(max_timeouts: 5, max_conn_resets: 5)
+        loop do
+          begin
+            return yield
+          rescue Net::OpenTimeout
+            raise if max_timeouts <= 0
+            max_timeouts -= 1
+          rescue Errno::ECONNRESET
+            raise if max_conn_resets <= 0
+            max_conn_resets -= 1
```
XaF, line 141: Could it make sense to:
1/ Add a slight delay before retry through `sleep`, maybe an increasing delay between each retry (1, 2, 4, etc.) up to a maximum delay if we don't want to wait too long
2/ Add an output message along with the delay, so deployments pipelines that monitor outputs do not reach a timeout before the end of the deployment

lib/record_store/provider/dnsimple.rb
```diff
@@ -14,19 +14,23 @@ def supports_alias?
...
       def retrieve_current_records(zone:, stdout: $stdout)
-        session.zones.all_records(account_id, zone).data.map do |record|
-          begin
-            build_from_api(record, zone)
-          rescue StandardError
-            stdout.puts "Cannot build record: #{record}"
-            raise
-          end
-        end.compact
+        retry_on_connection_errors do
```
XaF, line 17: Why put the retries in the providers instead of in the place where we call the providers ?
Would be a single change that affects all ?

cttttt: - For this PR, I only want to patch operations that are already idempotent by design, deferring the effort of making mutable operations idempotent until it'll provide us value.  Patching things where provider methods are called would make it more difficult to guarantee that errors are originating from idempotent operations.  In a future state where all provider methods are retryable, we could move this retry code further down the call stack...even to the command line class. 
- Unit testing is easier this way as I need to test smaller units.  In fact, my original thought was to patch even closer to Net::HTTP (retrying on both timeouts and connection resets based on the request method, sorta like- or piggybacking on-how Net::HTTP's built in retry mechanism works), but besides some monkey patching in this repo, this code that uses Net::HTTP actually exists in DNSimple and NS1 gems.  I'm thinking about perhaps submitting PRs in both repos.  This could make this patch obsolete in the future as these gems would just be resilient by design.  But this PR would still bring value in the meantime.

lib/record_store/provider.rb
cttttt: 1) Yes.  Particularly after the connection reset.  I'll let it retry without a delay on OpenTimeout, though...as it had just timed out.  Burning through 5 timeouts could take some time, so the quicker, the better.
2) Will throw in a message.  This is a good catch.  I think I can model it off of what's done during rate limiting.

lib/record_store/provider.rb
```diff
@@ -128,6 +128,36 @@ def master_nameserver_for(zone_name)
...
+            $stderr.puts("Retrying in #{delay}s after a connection reset")
+            backoff_sleep(delay)
+            delay = [delay * backoff_multiplier, max_backoff].min
+          end
+        end
+      end
+
+      def backoff_sleep(delay)
+        sleep(delay)
+      end
```
XaF, line 160: Would be nice to be able to reuse the logic of https://github.com/Shopify/record_store/blob/ea938f03380e1d42f5622cabf4bf5650ca63e097/lib/record_store/provider/provider_utils/rate_limit.rb (maybe by creating a parent class `Waiter` for instance, and just having the `sleep_for` being inherited with a different message, so we could in this case add `ConnectionResetWaiter` that would show the `waiting #{delay}s to retry after a connection reset` message ?)

cttttt: Def something for a future PR for sure.

PR #178, opened 2020-08-27
escape hyphen in regex

sbfaulkner:
we were getting...
`zone.rb:270: warning: character class has '-' without escape: /^([a-zA-Z0-9-_]+\.[a-zA-Z0-9-_])\.api\.shopify\.io\.$/`

the regex does in fact have a small bug, although I can't reproduce the warning

this PR fixes the regex(es) by escaping the standalone hyphens


sbfaulkner:
they may accept it, but definitely not technically valid... it should be escaped or be the first or last character of the regex

Issue #179, opened 2020-08-31
Always request JSON from provider APIs, or handle HTML error response bodies


ryandv:
It appears that (at least) NS1 may transiently return HTML response bodies with a 503 status code. We can't parse these as JSON:

```
$ bin/record-store assert_empty_diff
pid: 32416
#<Thread:0x00005561b66ce9a0@/app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/zone.rb:55 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	23: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/zone.rb:59:in `block (2 levels) in modified'
	22: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/zone.rb:83:in `unchanged?'
	21: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/zone.rb:76:in `build_changesets'
	20: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/zone.rb:76:in `map'
	19: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/zone.rb:77:in `block in build_changesets'
	18: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/changeset.rb:40:in `build_from'
	17: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/provider.rb:61:in `build_zone'
	16: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/provider/ns1.rb:56:in `retrieve_current_records'
	15: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/provider/ns1.rb:72:in `records_for_zone'
	14: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/provider.rb:150:in `retry_on_connection_errors'
	13: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/provider.rb:150:in `loop'
	12: from /app/data/bundler/ruby/2.6.0/gems/record_store-6.2.1/lib/record_store/provider.rb:152:in `block in retry_on_connection_errors'
[...]

PR #181, opened 2020-09-21
Handle NS1 Unparseable Responses

meganniu:
### Problem
NS1 returns an HTML body with a 503 status (service unavailable). `NS1:Client` tries to parse the HTML body as a JSON, which throws an uncaught error.

Related to #179 

### Solution
Create a new error type `RecordStore::Provider::ProviderUnavailableError` which is raised when NS1 returns a 503 status. If the error is raised within `RecordStore::Provider#retry_on_connection_errors`, then the connection will be retried several times.

### DIFF
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -158,8 +158,7 @@ module RecordStore
             raise if max_retries <= 0
             max_retries -= 1
 
-            waiter.message = "Waiting to retry after receiving an unparseable response"
-            waiter.wait
+            waiter.wait("Waiting to retry after receiving an unparseable response")
           rescue Net::OpenTimeout, Errno::ETIMEDOUT
             raise if max_timeouts <= 0
             max_timeouts -= 1
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -3,7 +3,7 @@ require 'resolv'
 module RecordStore
   class Provider
     class Error < StandardError; end
-    class UnavailableError < Error; end
+    class UnparseableBodyError < Error; end
 
     class << self
       def provider_for(object)
@@ -154,10 +154,11 @@ module RecordStore
         loop do
           begin
             return yield
-          rescue UnavailableError
+          rescue UnparseableBodyError
             raise if max_retries <= 0
             max_retries -= 1
 
+            waiter.message = "Waiting to retry since provider is unavailable"
             waiter.wait
           rescue Net::OpenTimeout, Errno::ETIMEDOUT
             raise if max_timeouts <= 0
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -3,7 +3,7 @@ require 'resolv'
 module RecordStore
   class Provider
     class Error < StandardError; end
-    class ProviderUnavailableError < Error; end
+    class UnavailableError < Error; end
 
     class << self
       def provider_for(object)
@@ -154,7 +154,7 @@ module RecordStore
         loop do
           begin
             return yield
-          rescue ProviderUnavailableError
+          rescue UnavailableError
             raise if max_retries <= 0
             max_retries -= 1
 
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -2,6 +2,9 @@ require 'resolv'
 
 module RecordStore
   class Provider
+    class Error < StandardError; end
+    class ProviderUnavailableError < Error; end
+
     class << self
       def provider_for(object)
         ns_server =
@@ -136,6 +139,7 @@ module RecordStore
       def retry_on_connection_errors(
         max_timeouts: 5,
         max_conn_resets: 5,
+        max_retries: 5,
         delay: 1,
         backoff_multiplier: 2,
         max_backoff: 10
@@ -150,6 +154,11 @@ module RecordStore
         loop do
           begin
             return yield
+          rescue ProviderUnavailableError
+            raise if max_retries <= 0
+            max_retries -= 1
+
+            waiter.wait
           rescue Net::OpenTimeout, Errno::ETIMEDOUT
             raise if max_timeouts <= 0
             max_timeouts -= 1
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -145,7 +145,7 @@ module RecordStore
         max_backoff: 10
       )
         waiter = BackoffWaiter.new(
-          "Waiting to retry after a connection reset",
+          'Waiting to retry after a connection reset',
           initial_delay: delay,
           multiplier: backoff_multiplier,
           max_delay: max_backoff,
@@ -158,12 +158,12 @@ module RecordStore
             raise if max_retries <= 0
             max_retries -= 1
 
-            waiter.wait("Waiting to retry after receiving an unparseable response")
+            waiter.wait(message: 'Waiting to retry after receiving an unparseable response')
           rescue Net::OpenTimeout, Errno::ETIMEDOUT
             raise if max_timeouts <= 0
             max_timeouts -= 1
 
-            $stderr.puts("Retrying after a connection timeout")
+            $stderr.puts('Retrying after a connection timeout')
           rescue Errno::ECONNRESET
             raise if max_conn_resets <= 0
             max_conn_resets -= 1
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -158,7 +158,7 @@ module RecordStore
             raise if max_retries <= 0
             max_retries -= 1
 
-            waiter.message = "Waiting to retry since provider is unavailable"
+            waiter.message = "Waiting to retry after receiving an unparseable response"
             waiter.wait
           rescue Net::OpenTimeout, Errno::ETIMEDOUT
             raise if max_timeouts <= 0
--- a/lib/record_store/provider/ns1/client.rb
+++ b/lib/record_store/provider/ns1/client.rb
@@ -49,7 +49,7 @@ module RecordStore
 
       def raise_error(result)
         if Net::HTTPResponse::CODE_TO_OBJ[result.status.to_s] == Net::HTTPServiceUnavailable
-          raise RecordStore::Provider::ProviderUnavailableError, result.to_s
+          raise RecordStore::Provider::UnavailableError, result.to_s
         end
         raise RecordStore::Provider::Error, result.to_s
       end
--- a/lib/record_store/provider/ns1/client.rb
+++ b/lib/record_store/provider/ns1/client.rb
@@ -48,7 +48,7 @@ module RecordStore
       private
 
       def raise_if_error!(result)
-        return nil unless result.is_a?(NS1::Response::Error)
+        return unless result.is_a?(NS1::Response::Error)
         if result.is_a?(NS1::Response::UnparsableBodyError)
           raise RecordStore::Provider::UnparseableBodyError, result.to_s
         end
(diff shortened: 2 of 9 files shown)


lib/record_store/provider.rb
```diff
@@ -150,6 +154,11 @@ def retry_on_connection_errors(
         loop do
           begin
             return yield
+          rescue ProviderUnavailableError
+            raise if max_retries <= 0
+            max_retries -= 1
+
+            waiter.wait
```
XaF, line 161: You probably want to change the message line 148 to encompass both connection reset & provider unavailable. Currently, debug might not be easy otherwise!

lib/record_store/provider/ns1/client.rb
```diff
@@ -1,46 +1,58 @@
...
+        raise_error(result) if result.is_a?(NS1::Response::Error)
         nil
       end
+
+      private
+
+      def raise_error(result)
+        if Net::HTTPResponse::CODE_TO_OBJ[result.status.to_s] == Net::HTTPServiceUnavailable
+          raise RecordStore::Provider::ProviderUnavailableError, result.to_s
+        end
```
XaF: This is nice! To iterate on this, I'm wondering if we could maybe have an `UnparsableError` type also that would be raised for every error type that has content not parsable through JSON ? It would be a matter of trying to parse the error content and raise if not possible. That way we do not have to add exceptions for every type of error, and an unparsable error should always be a service issue and not an API error, so we know it's worth retrying!

lib/record_store/provider/ns1/patch_api_header.rb
```diff
@@ -18,6 +18,10 @@ def process_response(response)
         rate_limit.wait(sleep_time)
       end
 
+      if Net::HTTPResponse::CODE_TO_OBJ[response.code] == Net::HTTPServiceUnavailable
+        return NS1::Response::Error.new({}, response.code.to_i)
+      end
```
XaF, line 23: I would suggest replacing this by a `begin..end` around the `JSON.parse` to raise the `UnparsableError` I talked about above. That way the response code can stay the same. Maybe the `UnparsableError` could be part of the patch here (`class NS1::Response::UnparsableError < NS1::Response::Error; end`) ?

lib/record_store/provider.rb
meganniu: Good catch!

lib/record_store/provider/ns1/patch_api_header.rb
meganniu: That's really clean, thanks for the suggestion!

test/providers/ns1_connection_errors_test.rb
```diff
@@ -100,4 +100,20 @@ def test_zones_raises_after_too_many_conn_resets
       @ns1.zones
     end
   end
+
+  def test_zones_raises_after_too_many_unparseable_responses
+    BackoffWaiter.any_instance.stubs(:wait)
+    BackoffWaiter.any_instance.expects(:wait).times(5)
+
+    stub_request(:get, "https://api.nsone.net/v1/zones")
```
ryandv, line 108: Nice! Very clean way of testing this without messy fixtures/"cassettes".

lib/record_store/provider/ns1/client.rb
```diff
@@ -1,46 +1,58 @@
...
       def delete_record(zone:, fqdn:, type:)
         result = super(zone, fqdn, type)
-        raise(Error, result.to_s) if result.is_a?(NS1::Response::Error)
+        raise_error(result) if result.is_a?(NS1::Response::Error)
         nil
       end
+
+      private
+
+      def raise_error(result)
```
ryandv, line 50: Loving the error handling approach here! 💯

lib/record_store/provider/ns1/patch_api_header.rb
```diff
@@ -18,15 +25,17 @@ def process_response(response)
...
 
-      body = JSON.parse(response.body)
-      case response
-      when Net::HTTPOK
-        NS1::Response::Success.new(body, response.code.to_i)
-      else
-        NS1::Response::Error.new(body, response.code.to_i)
+      begin
+        body = JSON.parse(response.body)
+        case response
```
ryandv, line 30: Totally optional because this works:

What do you think of extending this switch/case statement to handle the `Net::HTTPServiceUnavailable` status code, and raise the same `UnparsableBodyError`?

Just noticing [here](https://github.com/Shopify/record_store/pull/181/files#diff-5f6190009b4d6922fcd1b3bf4862de14R157-R161) that the debug messages indicate that we're retrying because the service is unavailable, yet the error we're handling is unparseable JSON.

Perhaps `UnparsableBodyError` is really `ServiceUnavailableError`?

ryandv (approved):
Looks great! Not much for me to comment on here, and all optional =)

:shipit:

lib/record_store/provider/ns1/patch_api_header.rb
XaF: I'm the one that suggested moving to `UnparsableBodyError` since it allows to consider something is wrong, but that the error code is not necessarily `ServiceUnavailable`. This avoids requiring to handle every little status code that would leave us with a body that's not JSON, and handle the fact that if the body is not parsable, we will wait and retry.

Maybe we want to have a clearer error message when it happens though. Would require create a separate waiter though, or simply update the waiter class's `wait` method to take an optional `message` parameter that, when provided, goes to replace the default wait message. That way we could simply call `waiter.wait(message: 'Unable to parse response body')` for instance.

lib/record_store/provider.rb
```diff
@@ -150,6 +154,12 @@ def retry_on_connection_errors(
         loop do
           begin
             return yield
+          rescue UnparseableBodyError
+            raise if max_retries <= 0
+            max_retries -= 1
+
+            waiter.message = "Waiting to retry since provider is unavailable"
```
XaF, line 161: As I suggested in https://github.com/Shopify/record_store/pull/181#discussion_r492919001 you might want instead to update the class to support an optional `message` parameter when calling the `wait` method.

Here, that line changes the message which will then never be changed again depending on the error (we could alternate between the two kinds of errors for instance), while allowing a temporary parameter to override it "just this time" would make things better I feel!

lib/record_store/provider/ns1/client.rb
```diff
@@ -1,46 +1,58 @@
...
-
     class Client < ::NS1::Client
       def initialize(api_key:)
         super(api_key)
       end
 
       def zones
-        super
+        zones = super
+        raise_error(zones) if zones.is_a?(NS1::Response::Error)
```
XaF, line 13: May I suggest updating your `raise_error` into `raise_if_error!` and doing `return unless param.is_a?(NS1::Response::Error)` as first line of that method ? That would allow to avoid all the repeated ifs here.

lib/record_store/provider/ns1/client.rb
```diff
@@ -1,46 +1,58 @@
...
+        zone = super(name)
+        raise_error(zone) if zone.is_a?(NS1::Response::Error)
+        zone
       end
 
       def record(zone:, fqdn:, type:, must_exist: false)
         result = super(zone, fqdn, type)
-        raise(Error, result.to_s) if must_exist && result.is_a?(NS1::Response::Error)
+        raise_error(result) if must_exist && result.is_a?(NS1::Response::Error)
         return nil if result.is_a?(NS1::Response::Error)
```
XaF, line 26: We can avoid checking twice the same thing here:

```suggestion
        if result.is_a?(NS1::Response::Error)
          raise_error(result) if must_exist
          return nil
        end
```

XaF:
^ Added a few more comments "out of a review" by mistake. Sorry for all the notifications!

lib/record_store/provider/ns1/patch_api_header.rb
meganniu: Whoops, thanks for catching the old message @ryandv 
I think `BackoffWaiter` has a setter method for `message`, so I can use that :)

XaF: > I think `BackoffWaiter` has a setter method for `message`, so I can use that :)

It does but then you'll have to use it everywhere you use the backoff waiter! Hence my suggestion of allowing "on-use override" of that message. That being said. doing a `waiter.message = "xxxx"` just before each call to `waiter.wait` will work. Just not as neat to read IMO, but not that important!

meganniu: Missed that comment, sorry! Agreed, on-use message setting would be better.

lib/record_store/provider/ns1/client.rb
```diff
@@ -10,46 +10,45 @@ def initialize(api_key:)
...
-        raise_error(result) if result.is_a?(NS1::Response::Error)
+        raise_if_error!(result)
         nil
       end
 
       private
 
-      def raise_error(result)
+      def raise_if_error!(result)
+        return nil unless result.is_a?(NS1::Response::Error)
```
XaF, line 51: `nil` is not really needed here

```suggestion
        return unless result.is_a?(NS1::Response::Error)
```

lib/record_store/provider/provider_utils/waiter.rb
```diff
@@ -5,7 +5,7 @@ def initialize(message = nil)
 
   attr_accessor :message
 
-  def wait(sleep_time)
+  def wait(sleep_time, message = @message)
```
XaF, line 8: I would recommend a keyword argument instead of positional just to make sure that we know what we're doing:
```suggestion
  def wait(sleep_time, message: @message)
```

That'll require to explicitly mention `message:` when calling, but makes it clearer (at least to me) what we are doing.

<details><summary>One thing I'm wondering about Ruby is if it's evaluating the default values of methods when the method is called or when the class is instantiated. So I tested and this seems to be working. <em>(Click to expand)</em></summary>
<p>

```ruby
irb(main):001:0> class Test
irb(main):002:1>   def initialize
irb(main):003:2>     @val = 2
irb(main):004:2>   end
irb(main):005:1>   attr_accessor :val
irb(main):006:1>   def test(val: @val)
irb(main):007:2>     puts "val is #{val}; @val is #{@val}"
irb(main):008:2>   end
irb(main):009:1> end
=> :test
irb(main):010:0> t = Test.new
=> #<Test:0x00007fa8b0890250 @val=2>
irb(main):011:0> t.test
val is 2; @val is 2
=> nil
irb(main):012:0> t.test(val: 3)
val is 3; @val is 2
=> nil
irb(main):013:0> t.test
val is 2; @val is 2
=> nil
irb(main):014:0> t.val = 4
=> 4
irb(main):015:0> t.test
val is 4; @val is 4
=> nil
irb(main):016:0> t.test(val: 3)
val is 3; @val is 4
=> nil
irb(main):017:0>
```

</p>
</details>

lib/record_store/provider.rb
```diff
@@ -158,8 +158,7 @@ def retry_on_connection_errors(
             raise if max_retries <= 0
             max_retries -= 1
 
-            waiter.message = "Waiting to retry after receiving an unparseable response"
-            waiter.wait
+            waiter.wait("Waiting to retry after receiving an unparseable response")
```
XaF, line 161: Recommending to use single quotes unless needing interpolations:

```suggestion
            waiter.wait('Waiting to retry after receiving an unparseable response')
```

XaF (approved):
This looks great @meganniu !
Just a last few comments for you to decide what to do with. I'm fine with what we have :)

lib/record_store/provider/provider_utils/waiter.rb
meganniu: Thanks for the suggestion and for investigating! Using a keyword argument sounds better to me too.

PR #185, opened 2020-11-30
Add support for templating implicit records in a zone

David-Tamrazov:
This PR adds support for injecting implicit records into a zone based on a template provided by the consumer. When building its record set, each `Zone` will now loop through every template in its config and, for each template, inject any records it creates based on criteria defined within the template itself.

This allows consumers to define their own criteria for:
- which records should be used for template record generation
- the format of the template record and what data should go into it
- what sort of existing record conflicts should prevent the template-generated record from being injected

TO DO:
- generate a new release with templating support


bin/console
```diff
@@ -6,6 +6,7 @@ require 'record_store'
 
 RecordStore.zones_path = File.expand_path('../../dev/zones', __FILE__)
 RecordStore.config_path = File.expand_path('../../dev/config.yml', __FILE__)
+RecordStore.zones_path = File.expand_path('../../dev/templates', __FILE__)
```
XaF, line 9:
```suggestion
RecordStore.templates_path = File.expand_path('../../dev/templates', __FILE__)
```

XaF: Do we ignore the `zones_path` and `templates_path` parameters in the `config.yml` when using the console ? (I guess yes since it's just for testing ?)

lib/record_store/zone.rb
```diff
@@ -189,7 +189,9 @@ def build_authority(authority)
     end
 
     def build_records(records)
-      records.map { |record| Record.build_from_yaml_definition(record) }
+      all_records = records.map { |record| Record.build_from_yaml_definition(record) }
+      config.templates.each { |template| template.inject_records!(all_records: all_records) }
```
XaF, line 193: Any reason to use a named parameter for that method ?

lib/record_store/zone/config.rb
```diff
@@ -3,15 +3,17 @@ class Zone
     class Config
       include ActiveModel::Validations
 
-      attr_reader :ignore_patterns, :providers, :supports_alias
+      attr_reader :ignore_patterns, :providers, :supports_alias, :templates
 
       validate :validate_zone_config
 
-      def initialize(ignore_patterns: [], providers: nil, supports_alias: nil)
+      def initialize(ignore_patterns: [], providers: nil, supports_alias: nil, templates: [])
```
XaF, line 10: I would perhaps have named this with a more explicit name to make sure that we know what those templates are for: `inject_records_templates` or `implicit_records_templates` for instance

lib/record_store/zone/config/template.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+module RecordStore
+  class Zone
+    class Config
+      class TemplateContext
```
XaF, line 6: Neat!

lib/record_store/zone/config/template.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        class << self
+          def from_file(filename:)
+            filepath = template_filepath_for(filename: filename)
+
+            raise InvalidFilepathError, "Could not find template '#{filename}' at location #{filepath}\n"\
+              unless File.exists?(filepath)
+
+            template_file = File.read(filepath)
+            filters_for_records_to_template = YAML.load(template_file).deep_symbolize_keys[:each_record]
```
XaF, line 37: Could the exception be done as a `rescue` from a `read` error ? This would give the exact error, including if there's an issue reading the file that might exist but not have the right permissions, for instance

We might also want to rescue on the `YAML.load` error to handle the case where the file is not properly formatted in YAML.

lib/record_store/zone/config/template.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        def inject_records!(all_records:)
+          all_records
+            .select { |record| should_template?(record: record) }
+            .map { |record| template_record_for(record: record, all_records: all_records) }
+            .each do|template_record|
+              if should_inject?(template_record: template_record, all_records: all_records)
+                all_records << Record.build_from_yaml_definition(template_record[:injected_record].first)
+              end
+            end
```
XaF, line 62: So we want to keep that `if` in the loop to avoid injecting twice the same record, for instance, if it matches two 'A' records. But we can avoid the extra block: (nitpicking though)

```suggestion
            .each do |template_record|
              next unless should_inject?(template_record: template_record, all_records: all_records)
              all_records << Record.build_from_yaml_definition(template_record[:injected_record].first)
            end
```

lib/record_store/zone/config/template.rb
```diff
@@ -0,0 +1,97 @@
...
+            record.public_send(key) == value
+          end
+        end
+
+        def template_record_for(record:, all_records:)
+          context = TemplateContext.build(record: record, all_records: all_records)
+
+          YAML.load(
+            template.result(context.get_binding)
+          ).deep_symbolize_keys
```
XaF, line 92: Should we handle exceptions here?

template/templates/template_example.yml.erb
```diff
@@ -0,0 +1,12 @@
+each_record:
+  - type: A
+    fqdn: abc.123.com.
+  - type: TXT
+conflict_with:
+  - type: TXT
+    fqdn: <%= "#{record.fqdn}.more.domain.com" %>
```
XaF, line 7:
```suggestion
each_record:
- type: A
  fqdn: abc.123.com.
- type: TXT
conflict_with:
- type: TXT
  fqdn: <%= "#{record.fqdn}.more.domain.com" %>
```

template/templates/template_example.yml.erb
```diff
@@ -0,0 +1,12 @@
...
+injected_record:
+- type: CNAME
+  ttl: 3600
+  fqdn: <%= "#{record.fqdn}.added.information.com." %>
```
XaF, line 11: Would work too: (nitpicking)

```suggestion
  fqdn: <%= record.fqdn %>.added.information.com.
```

XaF (commented):
I know you're currently working on adding tests. Here's a first batch of comments 😀

bin/console
David-Tamrazov: whoops thanks for the save - i think i just miscommitted what i have locally. let me repush

lib/record_store/zone.rb
David-Tamrazov: I just prefer named parameters because they make multi-argument methods easier to read/use, so i use them out the gate to make refactoring interfaces easier down the road if I ever need to.

lib/record_store/zone/config/template.rb
David-Tamrazov: yeah we can use the base exception if we want, this just reformatted the info a bit. 

Will think about how to handle the yaml errors as well.

lib/record_store/zone/config/template.rb
David-Tamrazov: I like it!

lib/record_store/zone/config/template.rb
David-Tamrazov: will consider in wider error handling strategy for https://github.com/Shopify/record_store/pull/185#discussion_r532863494

lib/record_store/zone/config/template.rb
David-Tamrazov: in other places we dont handle these sorts of errors unless we want to add on something special to the reporting, or have error handling logic, which we dont in these cases, so i'm opting to just let errors bubble up naturally for filepath and yaml format problems

lib/record_store/zone/config/template.rb
David-Tamrazov: https://github.com/Shopify/record_store/pull/185#discussion_r532936550

lib/record_store/zone/config/template.rb
XaF: It totally makes sense! Thanks for checking things to arrive to that conclusion :)

lib/record_store/zone.rb
```diff
@@ -189,7 +189,9 @@ def build_authority(authority)
     end
 
     def build_records(records)
-      records.map { |record| Record.build_from_yaml_definition(record) }
+      all_records = records.map { |record| Record.build_from_yaml_definition(record) }
+      config.implicit_record_templates.each { |template| template.inject_records!(all_records: all_records) }
```
dschwar, line 193: It feels weird that you're modifying the original records here. Consider passing them in and having the template return the modified records

David-Tamrazov: we want to modify the original records so that each iteration of template injection does its work with all the records in the zone up to that point (i.e. with full context). otherwise we have to do a tonne of conflict resolution in this method to make sure all the records returned by the templates dont conflict with each other

dschwar: Why not something like
```
records = all_records
config.implicit_record_templates.each do |template|
  records = template.inject_records!(records)
end
```

dschwar (approved):

lib/record_store/zone.rb
David-Tamrazov: hmm ruby is pass by reference so that ^ does pretty much the same as whats here. would it be better if we create an empty array thats a duplicate of `all_records`, pass that array around between templates, and then return the cumulative array at the end, vs modifying the original?

test/fixtures/config/dummy/config.yml
```diff
@@ -4,3 +4,4 @@ zones:
   - empty.com
   - one-record.com
   - two-providers.com
+implicit_record_templates_path: templates/implicit_records/
```
XaF, line 7: nitpick and also english is not my first language, `implicit_record_templates` vs. `implicit_records_templates` ? (i.e. could it convey the message of a single record in the first case ? in which case, we might want to add that `s` ?)

lib/record_store/zone.rb
XaF: Since the original is generated using `map`, it's already a "new array" though.
What @dschwar offers is nice to read, but to have a method that works that way would, as you said, require to either `.dup` the array at the beginning, or return just the records to inject and merge them here.

That could be something like:
```suggestion
      config.implicit_record_templates.each { |template| all_records.push(*template.records_to_inject(current_records: all_records)) }
```

With `records_to_inject` returning either an empty list (nothing to inject) or a list containing new records to inject. That would also keep the `all_records` list up to date. What are your thoughts on this ?

Also, nitpick: I renamed the parameter as `current_records`. I feel that in the `record_store` we generally avoid named parameters unless they're necessary, but I don't mind much having an extra name here, I suggest however to name it `current_records` instead of `all_records` as it feels clearer about what it means. ("all records" is for me the final version, with everything injected, in some way)

test/zone_test.rb
```diff
@@ -519,6 +519,43 @@ def test_zone_validates_support_for_ptr_records
     end
   end
 
+  def test_implicit_record_injection_only_injects_for_matching_records
```
XaF, line 522: Should we add a test for `no injection if no template` ?

lib/record_store/zone.rb
David-Tamrazov: i don't have strong feelings here - if we feel its clearer to do it this way, I'm happy to make the change. this way still lets us delegate conflict resolution to the individual templates with all relevant context provided, which is what my main requirement would be, so i'm good with it.

test/fixtures/config/dummy/config.yml
David-Tamrazov: i thiiink `implicit_record_templates` is the correct version here since each template is a template for one (type of) implicit record

XaF: Though there could be many records injected by a single template ? `injected_record` is a list (and should probably be `injected_records` or `records_to_inject` ?)

lib/record_store/zone.rb
XaF: No strong feeling either. I'd say, since the change for this is minimal, and it might make things slightly clearer when reading (making it directly seen that we're changing the `all_records` list, instead of hidden in the method call), let's go for it!

XaF (approved):

test/fixtures/config/dummy/config.yml
David-Tamrazov: but each template is only for one record - how many times we do it is an implementation detail, but the template itself only specifies and generates the single record, i.e. the implicit record. grammatically `implicit_record_templates` makes more sense to me

PR #191, opened 2021-09-08
[WIP] Check new records have not been delegated when adding a new record

ingytron:
This PR adds a check which ensures that a record hasn't been delegated elsewhere, thus having no effect, when adding a new record to a zone.


lib/record_store/zone.rb
```diff
@@ -272,6 +273,16 @@ def validate_provider_can_handle_zone_records
...
 
+    def validate_zone_not_delegated_to_another_provider
+      ns_records = records.select { |record| record.is_a?(Record::NS) }
+
+      ns_records.each do |records|
+        unless record.contains("nsone.net.") || record.contains(".dnsimple.com.")
+          errors.add(:records, "Warning, a zone delegation exists for this domain, this record will have no effect")
+        end
+      end
+    end
```
XaF: So if I understand correctly, this:
- Selects all NS records in the file
- Checks that any of those NS records is delegated to something else than NS1 and DNSimple
- Returns an error if it's the case

There's a few issues there:
1. This will complain as soon as we add any NS record to the zone. I would delegate `a.example.com` and this would complain right away, even if there is no record being shadowed by the delegation
2. This considers that NS1 and DNSimple records are always fine, and thus would ignore `b.a.example.com` even if there is a delegation for `a.example.com`, thus allowing me to add the record in the wrong zone, and to be shadowed

The first element to consider here is that we don't really care about the provider. The way things are structured in the `record_store` allows us to consider that as long as there is a delegation, I can error on any kind of record shadowed by that delegation. If that delegation brings back to the same providers, then I will have a different zone file to store the information. Putting the records in the wrong file would have them being shadowed and not working.

We thus need to:
- Given a list of records `records`
- List all NS records `ns_records` from `records`, except for those that are for the current zone (`example.com`)
- For each record in `records`, check if the `fqdn` of that record ends with any of the `ns_records`'s `fqdn`
- If they do: there's an error as that record is shadowed, as the subzone has been delegated

Example of case that's an issue:

```
example.com:
  config:
    providers:
      - SomeProvider1
      - SomeProvider2
  records:
    - type: NS
       fqdn: example.com.
       ttl: 3600
       nsdname: ns1.something.com.
    - type: NS
       fqdn: example.com.
       ttl: 3600
       nsdname: ns2.something.com.

    - type: CNAME
       fqdn: b.example.com.
       ttl: 3600
       cname: cname.example.com.

    - type: NS
       fqdn: a.example.com.
       ttl: 3600
       nsdname: ns1.somethingelse.com.

    - type: CNAME.   ##### THIS RECORD IS SHADOWED
       fqdn: b.a.example.com.
       ttl: 3600
       cname: cname.example.com.
```

lib/record_store/zone.rb
```diff
@@ -272,6 +273,26 @@ def validate_provider_can_handle_zone_records
       end
     end
 
+    def validate_zone_record_not_shadowed
+      nameserver_records = records
```
oefd, line 277: bit of a nitpick but maybe `nameserver_fqdns` since this isn't `Record` objects but rather just fqdns in strings?

lib/record_store/zone.rb
```diff
@@ -272,6 +273,26 @@ def validate_provider_can_handle_zone_records
...
+        .select { |record| record.is_a?(Record::NS) && name != record.fqdn }
+        .map { |record| record.fqdn.delete_suffix(".") }
+        .uniq
+
+      nameserver_records.each do |ns_record|
+        selected_records = records.reject do |record|
+          record.is_a?(Record::NS) && \
+            record.fqdn.delete_suffix(".") == ns_record
+        end
+        selected_records.each do |record|
```
ryandv, line 287: total nitpick, lean into collection pipelines and inline this temp variable:

```suggestion
       records.reject do |record|
          record.is_a?(Record::NS) && \
            record.fqdn.delete_suffix(".") == ns_record
       end.each do |record|
```

ryandv (approved):

lib/record_store/zone.rb
```diff
@@ -272,6 +273,26 @@ def validate_provider_can_handle_zone_records
...
+          normalized_record = record.fqdn.delete_suffix(".")
+          next unless normalized_record.end_with?(".#{ns_record}") || normalized_record == ns_record
```
ryandv, line 289: Could even slot a `.filter { ... }` into this collection pipeline perhaps, instead of the `next`

ryandv:
Oops, almost forgot. Please update the [CHANGELOG](https://github.com/Shopify/record_store/blob/master/CHANGELOG.md) and bump the gem version! (This can also take place in a subsequent PR)

XaF (approved):

lib/record_store/zone.rb
ingytron: I really love that but I'm going to skip it for readability

PR #193, opened 2021-09-20
account for race in mutex usage

oefd:
Had a fun error running tests just once - seems very intermittent:

```
  2) Error:
ZoneTest#test_modified_returns_all_zones_with_changes:
NoMethodError: undefined method `unchanged?' for nil:NilClass
    record_store/lib/record_store/zone.rb:66:in `block (2 levels) in modified'
```

I can't be 100% sure what I'm doing addresses the issue, but it seems likely since ignoring race conditions I'm not sure how else you'd end up with a nil `current_zone`.

Also just made it one lock/unlock action on the mutex

### DIFF
--- a/lib/record_store/zone.rb
+++ b/lib/record_store/zone.rb
@@ -61,8 +61,11 @@ module RecordStore
           Thread.new do
             current_zone = nil
             while zones.any?
-              mutex.synchronize { current_zone = zones.shift }
-              mutex.synchronize { modified_zones << current_zone } unless current_zone.unchanged?
+              mutex.synchronize do
+                next unless zones.any? # recheck zones since there's a race in the while
+                current_zone = zones.shift
+                modified_zones << current_zone unless current_zone.unchanged?
+              end
             end
           end
         end.each(&:join)


lib/record_store/zone.rb
```diff
@@ -61,8 +61,12 @@ def modified(verbose: false) # rubocop:disable Lint/UnusedMethodArgument
           Thread.new do
             current_zone = nil
             while zones.any?
-              mutex.synchronize { current_zone = zones.shift }
-              mutex.synchronize { modified_zones << current_zone } unless current_zone.unchanged?
+              mutex.synchronize do
+                if zones.any? # recheck zones since there's a race in the while
+                  current_zone = zones.shift
+                  modified_zones << current_zone unless current_zone.unchanged?
+                end
```
XaF, line 68:
```suggestion
                next unless zones.any? # recheck zones since there's a race in the while
                current_zone = zones.shift
                modified_zones << current_zone unless current_zone.unchanged?
```

XaF (approved):
I'm wondering if there was a reason why there was two mutex synchronization in there initially. But can't see why the new approach would break so...

XaF:
Ok figured-out the logic behind the broken-up mutex: `.unchanged?` takes time to verify. If we verify that while holding the mutex, no other process can do anything at all. We thus have a first access to get the zone in a secure manner, which makes sure that we don't have two threads working on the same zone. Then each thread can securely verify `.unchanged?` without holding the lock, and needs the lock back to make changes to the common variable if there is any change.

PR #194, opened 2021-09-20
restore broken up mutex syncs to allow change check outside critical section

oefd:
https://github.com/Shopify/record_store/pull/193#issuecomment-923095441

### DIFF
--- a/lib/record_store/zone.rb
+++ b/lib/record_store/zone.rb
@@ -61,10 +61,12 @@ module RecordStore
           Thread.new do
             current_zone = nil
             while zones.any?
-              mutex.synchronize do
-                next unless zones.any? # recheck zones since there's a race in the while
-                current_zone = zones.shift
-                modified_zones << current_zone unless current_zone.unchanged?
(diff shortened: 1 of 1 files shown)


lib/record_store/zone.rb
```diff
@@ -61,10 +61,12 @@ def modified(verbose: false) # rubocop:disable Lint/UnusedMethodArgument
           Thread.new do
             current_zone = nil
             while zones.any?
-              mutex.synchronize do
-                next unless zones.any? # recheck zones since there's a race in the while
-                current_zone = zones.shift
-                modified_zones << current_zone unless current_zone.unchanged?
+              mutex.synchronize { current_zone = zones.shift }
+              next if current_zone.nil? # account for the race between `zones.any?` and `zones.shift`
```
XaF, line 65: Realized that we don't need to check the loop on `zones.any?` again if this happens, right?

```suggestion
              mutex.synchronize { current_zone = zones.shift }
              break if current_zone.nil? # account for the race between `zones.any?` and `zones.shift`
```

XaF (approved):

PR #195, opened 2021-09-22
Bump gem version

ingytron:
This PR releases record_store gem version 6.5.3 to add support for shadowed record detection. 

Related to https://github.com/Shopify/record_store/pull/191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## 6.5.3
-- Adds check to detect shadowed records, for when a record being added to a zone in record-store will have no effect because it is shadowed by another record.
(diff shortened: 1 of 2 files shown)


mathu97 (approved):

oefd:
Shouldn't this be 6.6.0 given it's (arguably) a new feature?

Eh, I'm not dead set on interpreting it that way, just putting it out there.

XaF:
> Shouldn't this be 6.6.0 given it's (arguably) a new feature?
> 
> Eh, I'm not dead set on interpreting it that way, just putting it out there.

I also wondered, and realized it's not really a new feature of the `record_store`, but more a new error handling, that was already an error before, but wasn't handled properly, so.. sort of a bugfix ? (last minor was when we introduced implicit templates, which is an actual user feature)

oefd (approved):

PR #198, opened 2021-09-23
Update Rubocop config

ingytron:
Rubocop isn't working on master branch and it's because the config had changed but this repo is still configured to use the old config, resulting in a `404 "Not Found" while downloading remote config file ...` error. This PR updates it and makes rubocop runnable again.

Please note that this change generated a lot of `Correctable` offenses:
`61 files inspected, 1701 offenses detected, 1692 offenses auto-correctable`
For this, I added some custom cops to ignore a lot of the errors and then autocorrected the remaining 15. 

I'm definitely flexible when it comes to which cops to enable/disable but wanted to avoid changing too many files at once. 😄

### DIFF
--- a/lib/record_store.rb
+++ b/lib/record_store.rb
@@ -53,7 +53,7 @@ module RecordStore
     def zones_path
       @zones_path ||= Pathname.new(
         File.expand_path(config.fetch('zones_path'),
-        File.dirname(config_path)),
+          File.dirname(config_path)),
       ).realpath.to_s
     end
 
@@ -64,7 +64,7 @@ module RecordStore
     def implicit_records_templates_path
       @implicit_records_templates_path ||= Pathname.new(
         File.expand_path(config.fetch('implicit_records_templates_path'),
-        File.dirname(config_path)),
+          File.dirname(config_path)),
       ).realpath.to_s
     end
 
--- a/lib/record_store/provider.rb
+++ b/lib/record_store/provider.rb
@@ -3,6 +3,7 @@ require 'resolv'
 module RecordStore
   class Provider
     class Error < StandardError; end
+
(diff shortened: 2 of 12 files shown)


lib/record_store/provider.rb
```diff
@@ -152,24 +153,22 @@ def retry_on_connection_errors(
...
-            max_timeouts -= 1
-
-            $stderr.puts('Retrying after a connection timeout')
-          rescue Errno::ECONNRESET
-            raise if max_conn_resets <= 0
-            max_conn_resets -= 1
-
-            waiter.wait
-          end
+          return yield
```
ingytron, line 156: Rubocop detected redundant `begin` statements, and that appears to be consistent with the [Ruby Style Guide](https://rubystyle.guide/#no-BEGIN-blocks)

ryandv: This is a good cop to have on. :+1:

ryandv (approved):

XaF (approved):

lib/record_store.rb
```diff
@@ -53,7 +53,7 @@ def secrets_path
     def zones_path
       @zones_path ||= Pathname.new(
         File.expand_path(config.fetch('zones_path'),
-        File.dirname(config_path)),
```
XaF, line 56: Not totally sure how I feel about that indentation, since we're still in the same parenthesis... 🤷

ingytron: I see what you mean but if I change the indentation there it detects as an alignment offense so I'd need to disable the `[Layout/ArgumentAlignment](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/ArgumentAlignment)` cop to fix it. I suspect it's a Rubocop bug.

PR #200, opened 2022-02-08
it seems hash to kwargs is not implied any more

oefd:
I found a reference to ruby 2.7 deprecating automatic conversion of a hash to kwargs, but can't find a changelog reference that says it was fully removed at some point between 2.7 and 3.1

Regardless: I can say this is broken on ruby 3.1 because it wants kwargs, and doing this fixes it.


XaF (approved):

cttttt (approved):
lgtm

PR #235, opened 2024-04-04
prevent Alias conflicts

sbfaulkner:
I'm proposing that an ALIAS record shouldn't exist with the same FQDN as any A or AAAA record

It's possible I'm oversimplifying... happy to discuss.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## 6.8.0
+## 7.0.0
 - Validate that there are no A/AAAA record conflicts with ALIAS records
 
 ## 6.7.2
--- a/lib/record_store/version.rb
+++ b/lib/record_store/version.rb
@@ -1,3 +1,3 @@
 module RecordStore
-  VERSION = '6.8.0'.freeze
+  VERSION = '7.0.0'.freeze
 end


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 6.8.0
```
charlescng, line 3: Should it be a major version bump? It could be a breaking change.

test/zone_test.rb
```diff
@@ -197,6 +197,37 @@ def test_zone_with_cname_records
     )
   end
 
+  focus
```
charlescng, line 200: What's this for?

lib/record_store/zone.rb
```diff
@@ -265,6 +266,27 @@ def validate_cname_records_dont_point_to_root
...
+    ALIAS_CONFLICTS = [
+      "A",
+      "AAAA",
+    ]
+
+    def validate_alias_has_no_conflicts
+      alias_records = records.select { |record| record.is_a?(Record::ALIAS) }
+
+      alias_records.each do |alias_record|
+        records.each do |record|
```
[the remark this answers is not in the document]
sbfaulkner: this is validation for the zone, so it's the zone's records

CHANGELOG.md
sbfaulkner: I'm fine with that - in reality I don't think any provider should have accepted this in the first place, but as we observed recently it seems to be provider specific, so this validation makes it a defined behaviour

JLoppert (approved):

PR #244, opened 2024-08-02
Implement Initial Test Suite for Cloudflare Provider

tra0x:
The Cloudflare provider should be feature equivalent with the NS1 and DNSimple providers. As a first step towards achieving this equivalence, I have refactored the NS1 and DNSimple tests to also apply to the Cloudflare provider, which is currently pending implementation.

As the implementation progresses, tests may be added or removed based on the actual functionalities and requirements of the Cloudflare API.

All tests are set to be skipped at present.

### DIFF
--- a/test/cli_test.rb
+++ b/test/cli_test.rb
@@ -15,12 +15,13 @@ class CLITest < Minitest::Test
     Dir.mktmpdir do |dir|
       File.write(config_path = "#{dir}/config.yml", build_record_store_config)
       File.write("#{dir}/#{EJSON_PUBLIC_KEY}", EJSON_PRIVATE_KEY)
-      File.write(secrets_ejson_path = "#{dir}/secrets.#{ENV['CI'] ? 'ci' : 'dev'}.ejson", secrets_ejson = """
-{
-  \"_public_key\": \"#{EJSON_PUBLIC_KEY}\",
-  \"secret\": \"password\"
-}
-""")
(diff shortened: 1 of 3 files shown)


test/providers/cloudflare_test.rb
```diff
@@ -0,0 +1,525 @@
...
+      @cloudflare.apply_changeset(Changeset.new(
+        current_records: [],
+        desired_records: [record],
+        provider: @cloudflare,
+        zone: nil,
+      ))
+    end
+  end
+
+  def test_add_changset_missing_zone
```
anaclaudiar, line 343:
```suggestion
  def test_add_changeset_missing_zone
```

test/cli_test.rb
```diff
@@ -15,12 +15,13 @@ def test_secrets_uses_secrets_path_to_find_ejson_file
...
-  \"secret\": \"password\"
-}
-""")
+      File.write(secrets_ejson_path = "#{dir}/secrets.#{ENV['CI'] ? 'ci' : 'dev'}.ejson", secrets_ejson = <<~EJSON
+        {
+          "_public_key": "#{EJSON_PUBLIC_KEY}",
+          "secret": "password"
+        }
+      EJSON
+      )
```
sbfaulkner, line 24: I think this would be a lot more readable split into multiple lines assigning vars and then using

tra0x: This feels better just looking at it 😌

anaclaudiar (approved):
Nice, we will be doing test driven development!

PR #248, opened 2024-10-09
Cloudflare provider logic

tra0x:
Add Cloudflare provider logic to pass all tests. Tests were taken from NS1 and DNSimple test suites.

ALIAS record functionality is still to be implemented.


README.md
```diff
@@ -184,22 +184,22 @@ class Provider
   #
   # Arguments:
   # record - a kind of `Record`
-  def add(zone, record)
+  def add(record, zone)
```
tra0x, line 187: Update to match actual signature in code

test/providers/cloudflare_test.rb
```diff
@@ -104,381 +111,196 @@ def test_build_ptr_from_api
...
-      current_records: [],
-      desired_records: [record],
-      provider: @cloudflare,
-      zone: @zone_name,
-    ))
-    retrieved_records = @cloudflare.retrieve_current_records(zone: @zone_name)
-    assert_includes(retrieved_records, record)
-  end
-
-  def test_caa_record_retrieved_after_adding_record_changeset
```
anaclaudiar, line 402: Why these tests were removed?

lib/record_store/provider/cloudflare.rb
```diff
@@ -0,0 +1,183 @@
...
+
+      def secrets
+        super.fetch('cloudflare')
+      end
+
+      def build_api_body(record)
+        api_record = {
+          comment: '',
+          name: record.fqdn,
+          proxied: false,
```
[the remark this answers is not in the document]
tra0x: So I removed the `proxied: false,` line.

When a new record is created, cf will default to false.

Then I changed the update method to patch, and again do not set proxied. So an update will not modify proxied. That way if it is changed for a record in the dashboard, that change will persist.

For now, this lets us avoid managing settings in record_store. Only data directly part of the dns record is managed right now.

PR #249, opened 2024-10-10
Alias support in Cloudflare

tra0x:


test/providers/cloudflare_test.rb
```diff
@@ -160,6 +161,24 @@ def test_build_srv_from_api
     assert_equal(3600, record.ttl)
   end
 
+  def test_build_alias_from_api
```
anaclaudiar, line 164: Can you also add a test for the `build_api_body` ? Just to make sure something fails if anyone changes the implementation...

anaclaudiar (approved):
nice!

lib/record_store/provider/cloudflare.rb
```diff
@@ -123,7 +127,12 @@ def build_from_api(api_record)
         when 'A', 'AAAA'
           record.merge!(address: api_record['content'])
```
sbfaulkner, line 128: looks like we're not being consistent in use string or symbol to index the hash?
should this use `:content`?

lib/record_store/provider/cloudflare.rb
```diff
@@ -123,7 +127,12 @@ def build_from_api(api_record)
...
         when 'CNAME'
-          record.merge!(cname: api_record['content'])
+          if api_record['settings']['flatten_cname'] == false
```
sbfaulkner, line 130: as per my above comment we should try to be consistent in our sue of symbol or string to index the hash.. should this be `:settings` instead of `'settings'`?

sbfaulkner: others like `'content'` below too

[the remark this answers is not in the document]
tra0x: In my testing, the settings for a cname now always specifies either true or false

lib/record_store/provider/cloudflare.rb
```diff
@@ -82,74 +80,84 @@ def secrets
...
-          record.merge!(address: api_record['content'])
+          record.merge!(address: api_response['content'])
         when 'CNAME'
-          record.merge!(cname: api_record['content'])
+          if api_response['settings']['flatten_cname'] == false
+            record.merge!(cname: api_response['content'])
+          elsif api_response['settings']['flatten_cname'] == true
+            record_type = 'ALIAS'
+            record.merge!(alias: api_response['content'])
+          end
```
sbfaulkner, line 135:
```suggestion
          if api_response.dig('settings', 'flatten_cname')
            record_type = 'ALIAS'
            record.merge!(alias: api_response['content'])
          else
            record.merge!(cname: api_response['content'])
          end
```

is `settings` guaranteed to be present?
maybe consider this as an alternative?

sbfaulkner (approved):

PR #251, opened 2024-10-16
Batch Update DNS Records in Cloudflare 

tra0x:
Implements `apply_changeset` in Cloudflare provider rather than parent class. Individual `add`, `remove` and `update` methods have all been combined as part of the batch operation in `apply_changeset`.


lib/record_store/provider/cloudflare.rb
```diff
@@ -36,44 +36,46 @@ def zones
...
+          case change.type
+          when :removal
+            stdout.puts "Removing #{change.record}..."
+            deletes << { id: change.record.id }
+          when :addition
+            stdout.puts "Creating #{change.record}..."
+            posts << build_api_body(change.record)
+          when :update
+            stdout.puts "Updating record with ID #{change.id} to #{change.record}..."
+            patches << build_api_body(change.record).merge(id: change.id)
```
sbfaulkner, line 55: is there a reason you're using patches instead of puts?
in this case am I right that either could be used since we're likely providing all fields anyways?

sbfaulkner (approved):

PR #260, opened 2024-11-04
Fix Changeset normalization

tra0x:
Normalization has been incorporated into the `Changeset` class, where diffs are calculated. Since providers are solely responsible for retrieving current records and do not participate in the diff calculations, it was not feasible to limit normalization to just the Cloudflare provider alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 8.0.2
+- Fix Changesets to be case-insensitive
+
 ## 8.0.1
 - Fix Cloudflare provider to handle paginated returns from API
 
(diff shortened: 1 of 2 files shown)


test/providers/cloudflare_test.rb
```diff
@@ -111,15 +111,16 @@ def test_build_ptr_from_api
...
+      "name" => "0.183.136.115.in-addr.arpa",
+      "content" => "tbnbz2ni.byoip.example.com",
       "ttl" => 3600
     }
 
     record = @cloudflare.send(:build_from_api, api_record)
 
     assert_kind_of(Record::PTR, record)
-    assert_equal('host.record-store-dns-tests.shopitest.com.', record.ptrdname)
+    assert_equal('tbnbz2ni.byoip.example.com.', record.ptrdname)
```
anaclaudiar, line 122: I was expecting to see one test with capital letters like TBnbz2ni... no?

tra0x: that would be this [test](https://github.com/Shopify/record_store/blob/c6fad2005b59ccaa0518a771c5197c4cfb124af6/test/providers/cloudflare_test.rb#L241)

PR #262, opened 2024-11-06
Normalize A and AAAA records

tra0x:


lib/record_store/record/a.rb
```diff
@@ -26,5 +30,11 @@ def valid_address?
     rescue IPAddr::InvalidAddressError
       errors.add(:address, 'is invalid')
     end
+
+    def normalize_address(value)
+      IPAddr.new(value).to_s
+    rescue IPAddr::InvalidAddressError
+      value
```
AaliaM, line 37: Is there a reason why we're returning value when we get an `InvalidAddressError` (instead of raising an error)

anaclaudiar: In that case, you could remove the 
```rescue IPAddr::InvalidAddressError
      value```
from the normalize function then...

File: lib/record_store/record/aaaa.rb [hunk 2]
```diff
@@ -26,5 +30,11 @@ def valid_address?
     rescue IPAddr::InvalidAddressError
       errors.add(:address, 'is invalid')
     end
+
+    def normalize_address(value)
```
anaclaudiar, line 34: Does normalize_address something we will want to call for all record types somehow? I was wondering if we could define it in the parent class Record, then override if necessary in the children classes?

anaclaudiar (approved):
2 comments that are not blocking...

lib/record_store/record/aaaa.rb
tra0x: Probably a good future improvement

PR #273, opened 2025-02-26
Fix Cloudflare TXT record escaping

tra0x:
The string `no-quote` will send to Cloudflare as `"no-quote"`.

The string `"quotes"` will send to Cloudflare as `"\"quotes\""`. With escaped literal quotes.

Verified to work even with json encode/decode escapes applied.


sbfaulkner (commented):
seems correct.

unit tests are a good start...
1. how do we verify the behaviour is really as required for Cloudflare?
2. how do we rollout to all the existing domains - and make sure we're not breaking anything?

